PR #196, opened 2020-02-06
Made disabling Authereum notifications optional

TomAFrench:
I've had some feedback wanting Authereum notifications when using Onboard. I've added an optional initialisation value to set this behaviour by copying those for Torus.

Note: This PR switches the default to allowing notifications. If you don't want this I can add a little logic to swap this.


aaronbarnardsound:
@TomAFrench This is great! Yeah I agree it should default to having notifications.

PR #210, opened 2020-02-19
fix: add "jsonrpc: '2.0'" to sendAsync methods

TripleSpeeder:
Having the jsonrpc version string is mandatory according to https://www.jsonrpc.org/specification.

Depending on the node the missing version string is problematic. Having Metamask connected to a local parity node resulted in onboard.js not detecting the network version. Also on the browser console you would see an error logged from Metamask:

```
code: -32600
message: "Unsupported JSON-RPC protocol version"
```

With this patch applied no error is logged in the console and the connection to the local Parity node is accepted.


aaronbarnardsound:
Awesome! Good pickup @TripleSpeeder!

PR #220, opened 2020-02-26
Fix trust deep link

amiromayer:
Seems like you are using some outdated version of Trust wallet deep-link which does not work. So added a quick fix for that.


amiromayer:
https://developer.trustwallet.com/deeplinking

aaronbarnardsound:
Thanks for this @amiromayer 😄

PR #226, opened 2020-03-04
Torus integration

chaitanyapotti:
- bump version
- add support for enabled verifiers

### DIFF
--- a/src/validation.ts
+++ b/src/validation.ts
@@ -532,6 +532,7 @@ export function validateWalletInit(
     enableLogging,
     loginMethod,
     showTorusButton,
+    enabledVerifiers,
     disableNotifications,
     ...otherParams
   } = walletInit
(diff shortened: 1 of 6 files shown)


src/validation.ts
```diff
@@ -655,6 +657,14 @@ export function validateWalletInit(
     name: 'walletInit.showTorusButton',
     value: showTorusButton,
     type: 'boolean',
+    optional: false
```
aaronbarnardsound: Hey @chaitanyapotti did you mean to make the `showTorusButton` parameter no longer optional? This would be a breaking change, so will take some before we could get this merged in it's current state. I would recommend leaving this as an optional parameter if it doesn't break anything on your end.

chaitanyapotti: My bad.! It should still be optional. I was testing and made it non-optional.
I reverted it now.

aaronbarnardsound: No worries! Will get this merged soon 👍

PR #305, opened 2020-04-15
bump torus

### LINKED ISSUE #219: Torus Build Error (Rollup Apps)
The Torus SDK currently errors when trying to build in a Rollup app.

Screen Shot 2020-02-21 at 11 16 18 AM

I have replicated this in a basic Rollup build using Torus Embed directly without Onboard.

@torusresearch Have you guys successfully ran Torus Embed in a Rollup app? If so, is there a specific config parameter or plugin that is necessary?

chaitanyapotti:
we changed the default export to cjs modules now.
I'm not sure how that'll impact the import now.
```
const { default: Torus }  = await import('')
```
Please check with a sample example before merging this.
This new version of torus-embed will work with rollup seamlessly and we've tested with the rollup-starter-app as well.

Fixed #219 

@aaronbarnardsound


aaronbarnardsound:
Thanks will give this a look over shortly 👍

Issue #412, opened 2020-08-06
Add standard fallback font


benjlevesque:
On my laptop (windows 10), on which I don't have the `Helvtica Neue` font, Onboard.js falls back to `Times New Roman`. 

As far as I know, `Helvetica Neue` isn't a standard font and will not be present on all devices, maybe consider adding a fallback to a more common font?

![image](https://user-images.githubusercontent.com/14175665/89519872-58749100-d7dd-11ea-9d13-43ac3239d463.png)

PR #413, opened 2020-08-06
Add fallback fonts

benjlevesque:
As described in #412, if the font is missing, the default font is rendered, which looks pretty bad. 

I took the fallbacks from https://www.cssfontstack.com/Helvetica.


benjlevesque:
rebased `develop`, I initially created this PR on `master`

aaronbarnardsound:
Awesome, thanks for this @benjlevesque! Will get this in to the next release 👍

PR #414, opened 2020-08-12
add MYKEY wallet

geraldchen890806:


src/modules/select/index.ts
```diff
@@ -92,6 +93,8 @@ function getModule(name: string): Promise<any> | undefined {
       return import('./wallets/imtoken')
     case 'unilogin':
       return import('./wallets/unilogin')
+    case 'MYKEY':
```
aaronbarnardsound, line 96:
```suggestion
    case 'mykey':
```
Could you please switch this to either all lowercase or camelCase to be consistent with the rest of the wallet initialization names.

src/modules/select/index.ts
```diff
@@ -13,7 +13,8 @@ const defaultWalletNames = [
   'opera',
   'operaTouch',
   'status',
-  'unilogin'
+  'unilogin',
+  'MYKEY'
```
aaronbarnardsound, line 17:
```suggestion
  'mykey'
```
Please make this Lowercase or camelCase to match other wallet names

PR #423, opened 2020-08-21
Add huobiwallet support

lnwanghao:
Add huobiwallet support


src/modules/select/wallets/huobiwallet.ts
```diff
@@ -0,0 +1,74 @@
...
+  return {
+    name: label || 'Huobi Wallet',
+    svg: svg || huobiwalletIcon,
+    wallet: async (helpers: Helpers) => {
+      const { getProviderName, getAddress, getNetwork, getBalance } = helpers
+      const huobiwalletProvider =
+        (window as any).ethereum ||
+        ((window as any).web3 && (window as any).web3.currentProvider)
+
+      const isHuobiWallet = getProviderName(huobiwalletProvider) === 'huobiwallet'
```
aaronbarnardsound, line 22:
```suggestion
      const isHuobiWallet = 
        getProviderName(huobiwalletProvider) === 'huobiwallet'
```

Formatting change to satisfy linting rules for line length.

src/utilities.ts
```diff
@@ -215,6 +215,11 @@ export function getProviderName(provider: any): string | undefined {
     return 'MYKEY'
   }
 
+  if (provider.isHuobiWallet) {
```
aaronbarnardsound, line 218:
```suggestion
  if (provider.isHbWallet) {
```

During my testing on iOS, this is where the boolean flag namespace was. Is this the same for all devices?

aaronbarnardsound (changes requested):
Requested a couple of changes :)

aaronbarnardsound (approved):

PR #424, opened 2020-08-25
Enhancement: Ledger Transport

### LINKED ISSUE #415: Ledger Transport
Onboard.js currently defaults to the U2F transport, but this transport is [broken for Windows 10 users](https://www.ledger.com/windows-10-update-sunsetting-u2f-tunnel-transport-for-ledger-devices/).
Instead of defaulting to the U2F transport, we can detect the user's browser/os combination when initializing the Ledger module and dynamically import the ideal transport for that combination. 

The possible scenarios and the corresponding transport: 
- OS: Windows 10, Browser: Firefox | IE -- Display error message informing the user to switch to a compatible browser
- OS: macOS, Browser: Safari | Firefox -- Use U2F transport
-  All other combinations default to using the WebUSB transport

aaronbarnardsound:
Fixes #415

### DIFF
--- a/src/modules/select/wallets/ledger.ts
+++ b/src/modules/select/wallets/ledger.ts
@@ -206,8 +206,10 @@ async function ledgerProvider(options: {
             disconnect()
           }
         },
-        error: console.log,
-        complete: console.log
+        error: (error: any) => {
+          throw new Error(error)
+        },
+        complete: () => {}
       }
 
       if (CustomLedgerTransport) {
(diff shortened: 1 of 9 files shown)


src/modules/select/wallets/ledger.ts
```diff
@@ -185,27 +200,54 @@ async function ledgerProvider(options: {
...
+        ) {
+          const { default: TransportU2F } = await import(
+            '@ledgerhq/hw-transport-u2f'
+          )
+
+          transport = await TransportU2F.create()
+        } else {
+          const { default: TransportWebUsb } = await import(
+            '@ledgerhq/hw-transport-webusb'
+          )
```
liamaharon, line 239: is this some cool voodoo to reduce initial page load time?

aaronbarnardsound: Haha yep, dynamic imports will only load the module when and if it is needed rather than all at once on initial load, reducing data usage and load time - https://medium.com/rollup/rollup-now-has-code-splitting-and-we-need-your-help-46defd901c82

PR #463, opened 2020-12-10
Release 1.17.0

dmarzzz:


aaronbarnardsound (approved):

src/modules/select/wallets/torus.ts
```diff
@@ -59,25 +61,27 @@ function torus(options: TorusOptions & { networkId: number }): WalletModule {
...
-            get: () => Promise.resolve(Number(instance.web3.version.network))
+            get: () => Promise.resolve(Number(networkId))
           },
           balance: {
             get: () =>
               new Promise(async (resolve, reject) => {
                 instance.web3.eth.getBalance(
-                  instance.web3.eth.accounts[0],
-                  (err: any, data: any) => {
+                  account,
```
chuckbergeron, line 78: Thanks for putting this fix together! When I tried it out I was still getting the `invalid address` error. I threw these lines right after the `new Promise` to return early and now the exception is gone. Instead of I get warnings.

```js
if (!account) {
  reject(`Error while checking Balance, wallet not yet connected`)
  return
}
```

PR #464, opened 2020-12-11
add AToken wallet

iRobin520:
Add the AToken wallet into onboard


dmarzzz (approved):
Tested on mobile using aToken wallet

Issue #466, opened 2020-12-15
polling interval


dmarzzz:
<img width="657" alt="Screen Shot 2020-12-15 at 4 56 52 PM" src="https://user-images.githubusercontent.com/22778355/102277636-9190e080-3ef6-11eb-9643-6dc7891a8a9e.png">

Wallet-connect and web3-provider-engine can take in a pollingInterval parameter but onboard has no way to pass one in at the moment.

Functionality that needs to be implemented: Add an optional (blockPollingInterval) parameter to initialization, add to validation and svelte store, and pass it into wallet-connect.ts and providerEngine.ts modules.

aaronbarnardsound:
Closed via #467

PR #467, opened 2020-12-15
blockPollingInterval add

dmarzzz:
In regards to #466

### DIFF
--- a/src/onboard.ts
+++ b/src/onboard.ts
@@ -49,7 +49,8 @@ function init(initialization: Initialization): API {
     networkId,
     darkMode,
     apiUrl,
-    hideBranding
+    hideBranding,
+    blockPollingInterval = 40000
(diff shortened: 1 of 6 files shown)


src/onboard.ts
```diff
@@ -49,7 +49,8 @@ function init(initialization: Initialization): API {
     networkId,
     darkMode,
     apiUrl,
-    hideBranding
+    hideBranding,
+    blockPollingInterval = 40000
```
aaronbarnardsound, line 53:
```suggestion
    blockPollingInterval = 4000
```

aaronbarnardsound (changes requested):
Looks good, just need to change that default `blockPollingInterval` to the correct value.

aaronbarnardsound (approved):

PR #470, opened 2020-12-21
Fix Torus integration

### LINKED ISSUE #461: Torus Web3 Type Error
Type error in 'onboard/src/modules/select/wallets/torus.ts(102,40)'. Exported web3 type from Torus is incorrect and does not define an index to select on in the web3.eth.accounts array. 

Specific Error : 

Element implicitly has an 'any' type because expression of type '0' can't be used to index type 'Accounts'.
Property '0' does not exist on type 'Accounts'

chaitanyapotti:
Fixes #461 

Do you have a sample dapp where i can test the integration.?
If not, i'll create one myself


dmarzzz (approved):
Tested with onboard/react-demo, all functionality working

PR #475, opened 2021-01-02
Updates GridPlus connector in anticipation of upcoming Lattice firmware release

alex-miller-0:
This commit updates GridPlus-related modules to prepare for what will be
breaking changes in firmware. The updated packages are themselves backwards
compatible so current users will see no difference in functionality.


alex-miller-0:
I went ahead and rebased against the `develop` branch.

taylorjdawson:
@alex-miller-0 Excellent, thank you!

PR #484, opened 2021-01-21
Feature/move ws init

dmarzzz:
Moves bnc-sdk init to where balance store is created.

### DIFF
--- a/src/stores.ts
+++ b/src/stores.ts
@@ -1,7 +1,7 @@
 import { getBlocknative, initializeBlocknative } from './services'
 import { writable, derived, get } from 'svelte/store'
 import { wait, makeCancelable, createInterval } from './utilities'
-import { validateWalletInterface, validateType } from './validation' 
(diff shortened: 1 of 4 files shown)


src/stores.ts
```diff
@@ -1,7 +1,7 @@
-import { getBlocknative } from './services'
+import { getBlocknative, initializeBlocknative } from './services'
 import { writable, derived, get } from 'svelte/store'
 import { wait, makeCancelable, createInterval } from './utilities'
-import { validateWalletInterface, validateType } from './validation'
+import { validateWalletInterface, validateType } from './validation'
```
taylorjdawson, line 4: nit: trailing space

taylorjdawson (approved):

src/stores.ts
dmarzzz: Thank you! Fixed

cmeisl (approved):

PR #487, opened 2021-01-28
Feature/move ws init

dmarzzz:
Move ws init to trigger on getBlocknative() call


src/stores.ts
```diff
@@ -309,14 +309,6 @@ function createBalanceStore(initialState: string | null): BalanceStore {
             currentBalance: get(balance)
```
taylorjdawson, line 309: @aaronbarnardsound Something that I found interesting is the amount of `get`'s that are used. What is the rule of thumb when deciding between: `$<store object>` and `get(<store object>)`? It seemed like from the Svelte docs that they tend to use `$` more than `get`.

taylorjdawson (approved):

src/stores.ts
aaronbarnardsound: @taylorjdawson Generally I would use `$` to subscribe as much as possible since it will auto unsubscribe for you. But that is only valid in `.svelte` files. So in regular TypeScript/JavaScript files it is sometimes necessary to use the `get` function. In the above case we could have made `balance` a dependency of the `derived` call (like we did with `address` and `network`) but I believe I didn't do that in this case as I didn't want to re-run the function when the balance changes, only when the address or network changes. Another pattern is to just subscribe to the `balance` store and assign it to a `let` variable in the upper scope that will always be up to date and can be referred to anywhere in the file. I would do this if that value is needed in multiple places throughout the file.

taylorjdawson: @aaronbarnardsound ha of course! 🤦  I forgot that that is special svelte syntax. Okay that makes sense now, thanks for the explanation! 👍

aaronbarnardsound (approved):
Nice one, I'm going to merge this in

PR #494, opened 2021-02-13
1.19.2-0.2.0: [enhancement] - Add Frame

floating:
Adds wallet option to connect to Frame natively


aaronbarnardsound:
Hey @floating I spent some time testing this and I couldn't get it to work in [our demo app that we use for testing](https://github.com/blocknative/react-demo). I added some commits that get it closer to being ready, but there are JSON errors when trying to request access to accounts and it also seems to need an rpc endpoint to be passed in. If you could spend some time to get it working in the react demo locally and then we can review from there. Thanks :)

floating:
Hey @aaronbarnardsound! 

I've tried it with `react-demo` and everything seems to be working fine. 

What error are you getting?

floating:
Hey @aaronbarnardsound, any updates here?

aaronbarnardsound:
Hey @floating the errors I was seeing was due to not running Frame (I had not looked in to how it works and assumed that it was a SDK style wallet). 

I modified the interface so that it has disconnection logic for when user's switch wallets to prevent memory leaks. I also modified the error handling to display the frame error message for more helpful messages when connecting. There is logic in the error handling now to handle the "Unexpected end of JSON input" error for when trying to connect without running Frame, feel free to modify the message that is displayed to the user in the connect modal. I set it to "Frame is not running". I f you are happy with all of that, then we are good to merge this in to the develop branch.

floating:
Looks good to me!

aaronbarnardsound (approved):

PR #502, opened 2021-03-01
1.19.2-0.1.0: [feature] network display name option

taylorjdawson:
- Adds optional `networkName` to the config to allow for users to custom the display name of unsupported chains


aaronbarnardsound (approved):
Looks good 👍

Issue #506, opened 2021-03-07
Clicking "Dismiss" in OnBoard modals should flush localstorage if set


swaylock:
Gonna be hard to provide context for this issue, but I'll try.

The way OnBoard is coded, it seems you must have multiple buttons in your wallet connection UI/UX, as exemplified by the demo application:

<img width="606" alt="onboard" src="https://user-images.githubusercontent.com/23582496/110226600-a9e51b80-7ea5-11eb-84a6-13e636b137c3.png">

In our UI, we are wanting to present something much simpler: A single button the user interacts with -- very similar to connection UI/UX on 1Inch exchange.

For the current state of the problem I'm running into, try the following scenario in a browser:

1. Disable the Authereum wallet if it's enabled on your browser (i use brave and never installed it so that's my starting point).
2. Navigate to the demo application https://reactdemo.blocknative.com/
3. Click select a wallet and choose Authereum. You should get a popup dialog and some sort of error message.
4. Refresh the page or close the tab and re-open the demo app.
5. The Authereum connection dialog will pop again. This is because the demo app has saved your selection to localstorage and trying to reload Authereum because you selected it. 
6. Now here's the crux of the problem. Unless you present the user with multiple buttons, for example, "Select another wallet" or "Reset wallet state" or something similar, they will be stuck with the Authereum wallet dialog continually popping up. Or, even if you can coordinate all the logic necessary to deal with automatically reconnecting based on previous selection into a single button, the user will still need to close the unwanted popup and then click the connect button again and hopefully it will all work out.

What I'm proposing is an improvement by providing a way to flush saved wallet selection from localstorage from the Onboard modal dialogs. I'm suggesting we use the  "Dismiss" button (perhaps it should be renamed to "Cancel"?)

![screenshot-reactdemo blocknative com-2021 03 06-18_39_27](https://user-images.githubusercontent.com/23582496/110227333-0ba88400-7eac-11eb-8702-e3332018e919.png)

PR coming shortly.

aaronbarnardsound:
Hey @swaylock, thanks for detailing your issue / use case and also creating a PR for it :)

We have tried to design Onboard to be as flexible as possible so that it can be used for various Onboarding flows but without having infinite configuration options that make it more complex and harder to understand. I think that caching the wallet selection and managing that cache can stay out of the Onboard library since it is up to the app to decide whether to implement this or not.

[...]

PR #507, opened 2021-03-07
Create new config option for selected wallet localstorage key.

swaylock:
Related to gh-506

So, first off, I have to confess I'm not clear on the security issues with this proposed change. My gut tells me there are none, but it should definitely be reviewed/considered. Anybody over there a JS security expert?

Anyhow, this PR adds a brand new optional configuration option to OnBoard called `selectedWalletKey`
If this config option is set, the core onboard code will be able to flush the localstorage if necessary.

Onboard initialization could then look something like:

```javascript
const localstorageWalletKey = "myAppSelectedWallet";
onboard = Onboard({
      dappId: process.env.VUE_APP_BLOCKNATIVE_API_KEY,
      networkId: Number(process.env.VUE_APP_ONBOARD_NETWORK_ID),
      selectedWalletKey: localstorageWalletKey,
      subscriptions: {...
```
Then in your implementation code, you can set & get the currently selected wallet as the demo application currently does. For example:

``` javascript
// Called any time page is loaded. If a wallet
  // has been previously selected, it initializes
  // and checks the wallet.
  async reconnectWallet() {
    const previouslySelectedWallet = window.localStorage.getItem(
      localstorageWalletKey
    );
    if (!previouslySelectedWallet) {
      return false;
    }

    if (previouslySelectedWallet && onboard) {
      const walletSelected = await onboard.walletSelect(
        previouslySelectedWallet
      );
      return walletSelected;
    }
  },
```

Finally, now when you on on the wallet check modal and you click "Dismiss" the previously selected wallet in localstorage will be unset so it acts more like a clean refresh of the UX. Similar to onboard.walletReset()


aaronbarnardsound:
Resolved via #506

PR #508, opened 2021-03-12
1.20.0-0.1.0: [feature]: Add Cobo Vault hardware wallet support

soralit:
Hi, this PR adds support for [Cobo Vault](https://cobo.com/hardware-wallet/cobo-vault)

This PR integrate with cobo vault eth-keyring witch supports the common ETH related functionality and usage with Cobo Vault hardware.


src/modules/check/connect.ts
```diff
@@ -29,7 +29,7 @@ function connect(options: WalletCheckCustomOptions = {}): WalletCheckModule {
       }
     }
 
-    if (!stateStore.address.get() && wallet && wallet.name) {
+    if (!stateStore.address.get() && wallet && wallet.name && wallet.name !== 'CoboVault') {
```
aaronbarnardsound, line 32: I would suggest that we take this clause out that checks for CoboVault. I can see why you don't want the connect modal to render as it looks a little weird underneath the qr scanner modal that pops up. I think this modal is important as it gives the user some context, especially if the user clicks cancel on the cobo scanner modal, then the connect modal is there explaining that this step is necessary to continue with the app and allows them to click "Try again" to show the qr modal again.

Maybe you could tweak your modal design a bit so that it is full screen and darkens the background which would block out the connect modal until the user clicks cancel? Check out the WalletConnect modal on our react demo for an example of what that could look like.

src/modules/select/wallets/cobovault.ts
```diff
@@ -0,0 +1,261 @@
...
+    if (!enabled) {
+      return []
+    }
+
+    if (addressList.length > 0 && !getMore) {
+      return addressList
+    }
+
+    try {
+      addressList = await keyring.addAccounts(5)
```
aaronbarnardsound, line 184:
```suggestion
      addressList = await keyring.addAccounts(addressList.length + 5)
```

What do you think about changing the above code so that when the user clicks the "load more" button on the accounts modal, it loads another 5 addresses just like our other hardware wallet modules do?

src/modules/select/wallets/cobovault.ts
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  function enable() {
+    enabled = true
+    return keyring.readKeyring().then(() => {
+      return getAccounts()
+    })
+  }
+
+  function addresses() {
```
aaronbarnardsound, line 155: This function can be removed since it isn't being used.

src/modules/select/wallets/cobovault.ts
```diff
@@ -0,0 +1,261 @@
...
+    customPath = false
+    dPath = path
+    return true
+  }
+
+  function isCustomPath() {
+    return customPath
+  }
+
+  function enable() {
```
aaronbarnardsound, line 148:
```suggestion
  function enable() {
    if (enabled) {
      return getAccounts()
    }
```

If a user has already connected their Cobo Vault, then adding this in will allow them switch accounts without needing to go through the connection process again.

aaronbarnardsound (changes requested):
Just a few small suggestions and changes :)

src/modules/check/connect.ts
soralit: Hi @aaronbarnardsound, thanks a lot for your feedback, initially I think the Cobo Vault does not have a `connect` step in our user journey, because the authorization is a default behavior when user want to sync his Cobo Vault to a watch-only wallet via QR codes. 
So it will looks a little weird that I sync my wallet QR codes but nothing happened, then I need to sync my QR codes again in another QR scanner modal. I think it is caused by some inconsistent business logic. 
And I think we can revert this change and modify Cobo Vault's `connect` modal by adding some help texts like "Please scan Cobo Vault Sync QR codes for authorizing" and adding other help texts to `accounts` modal like "Please sync Cobo Vault to get accounts" to make it clear.
What do you think? Any better suggestions?

src/modules/select/wallets/cobovault.ts
soralit: Thanks.

src/modules/select/wallets/cobovault.ts
soralit: Thanks

src/modules/check/derivation-path.ts
```diff
@@ -182,56 +182,58 @@ function derivationPath(
...
-              state.dPath || derivationPaths[wallet.name as string][0].path
-            try {
-              const validPath = await wallet.provider.setPath(
-                path,
-                state.showCustomInput
-              )
-
-              if (!validPath) {
-                state.error = `${path} is not a valid derivation path`
+      if (wallet.name !== 'CoboVault') {
```
soralit, line 185: Also, how do you like this change ? @aaronbarnardsound 
For Cobo Vault not supporting customize derivation path currently, this check looks a lit weird for us.

src/modules/check/connect.ts
aaronbarnardsound: @soralit My concern is not as much for showing the modal informing the user to connect, but what happens if the user clicks "close" on your QR modal without scanning their wallet, it gets stuck on this screen:

Screen Shot 2021-03-24 at 10 48 40 am

Whereas if you include the Onboard connect modal it displays this:

Screen Shot 2021-03-24 at 10 51 35 am

This modal allows the user to click "Try Again" button which will display the QR modal again. So i think it is best to make the suggested change above.

src/modules/check/derivation-path.ts
aaronbarnardsound, line 185:
```suggestion
      
```

Yeah I think we remove this if statement and do the below suggestion instead

src/modules/check/derivation-path.ts
```diff
@@ -182,56 +182,58 @@ function derivationPath(
...
+        return {
```
aaronbarnardsound, line 186:
```suggestion
        return (
          derivationPaths[wallet.name as string] && {
```

aaronbarnardsound (changes requested):

src/modules/check/connect.ts
soralit: 👍

soralit:
Hi @aaronbarnardsound , thanks for your review, I have resolved code issues mentioned above, would you like to do a review again when you have time? 😄

aaronbarnardsound (approved):

aaronbarnardsound (approved):

PR #514, opened 2021-03-24
1.20.0-0.0.1: [enhancement] - Install Wallet Copy

### LINKED ISSUE #513: Mobile Wallet Install Message
It can be confusing to users when they are on a mobile device and they are informed by the install modal to install a wallet that they already have. We should change the language so the it informs to "open" instead of install.

aaronbarnardsound:
closes #513

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -4,5 +4,6 @@ module.exports = {
   singleQuote: true,
   printWidth: 80,
   tabWidth: 2,
-  arrowParens: 'avoid'
+  arrowParens: 'avoid',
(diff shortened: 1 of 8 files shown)


.prettierrc.js
```diff
@@ -4,5 +4,6 @@ module.exports = {
   singleQuote: true,
   printWidth: 80,
   tabWidth: 2,
-  arrowParens: 'avoid'
+  arrowParens: 'avoid',
+  svelteSortOrder: 'scripts-styles-markup'
```
taylorjdawson, line 8: 😍

taylorjdawson (approved):

PR #520, opened 2021-04-05
1.21.0-0.0.2: [fix] - Adds unknown network name

taylorjdawson:


aaronbarnardsound:
- This PR fixes an issue where if a custom `networkId` and a `networkName` is passed in, and the user happens to be on a network that Onboard doesn't recognize, it will now display in the network wallet check modal that the user needs to switch from "unknown" network to the `networkName` that is passed is for the custom network. Previously it would inform the user to switch from the custom name to the custom name.

aaronbarnardsound (approved):

PR #527, opened 2021-04-12
1.22.0-0.2.0: [enhancement] - Filter Confirmed Status

### LINKED ISSUE #526: Filter to 'confirmed' status
When watching the wallet's current address for efficient balance queries, Onboard is using the `account` method from the Blocknative SDK which will subscribe to all transaction events for that address. Since we only need `confirmed` status events to trigger a balance update, we can use the filtering capability of the SDK to minimize events that are delivered to the client.

aaronbarnardsound:
- Filters SDK events to just receive confirmed status event for efficient balance updates

Closes #526

### DIFF
--- a/src/stores.ts
+++ b/src/stores.ts
@@ -1,5 +1,6 @@
-import { getBlocknative } from './services'
+import { Emitter } from 'bnc-sdk/dist/types/src/interfaces'
 import { writable, derived, get } from 'svelte/store'
+import { getBlocknative } from './services'
 import { wait, makeCancelable, createInterval } from './utilities'
 import { validateWalletInterface, validateType } from './validation'
 import {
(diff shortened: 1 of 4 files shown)


src/stores.ts
```diff
@@ -1,5 +1,6 @@
-import { getBlocknative } from './services'
+import { Emitter } from 'bnc-sdk/dist/types/src/interfaces'
```
taylorjdawson, line 1: @aaronbarnardsound is there a reason why this can't be exported directly from the module?

taylorjdawson (approved):

taylorjdawson:
@aaronbarnardsound I think you'll also need to update the version to `1.22.0-0.2.0`

src/stores.ts
aaronbarnardsound: @taylorjdawson We'd need to export them from the sdk module directly which would be a nice improvement for sure

PR #531, opened 2021-04-16
1.23.0-0.1.0: [feature] Adds optional terms agreement

taylorjdawson:
Adds optional terms agreement

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,0 +1,5 @@
+import { StorageKeys } from './interfaces'
+
+export const STORAGE_KEYS: StorageKeys = {
+  TERMS_AGREED: 'onboard.js:termsAgreed'
+}
--- a/src/modules/index.ts
+++ b/src/modules/index.ts
@@ -12,24 +12,32 @@ const defaultWalletExplanation = `Wallets are used to send, receive, and store d
 
 export default function initializeModules(
   networkId: number,
-  walletSelect: WalletSelectModuleOptions | undefined,
+  walletSelectOptions: WalletSelectModuleOptions | undefined,
   walletCheck: Array<WalletCheckModule | WalletCheckInit> | undefined,
   isMobile: boolean
 ) {
   const wallets = select(
-    walletSelect && walletSelect.wallets,
+    walletSelectOptions && walletSelectOptions.wallets,
     networkId,
     isMobile
   )
 
+  const {
+    heading = defaultHeading,
+    description = defaultDescription,
+    explanation = defaultWalletExplanation,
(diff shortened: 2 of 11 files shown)


src/constants.ts
```diff
@@ -0,0 +1,5 @@
+import { StorageKeys } from './interfaces'
+
+export const STORAGE_KEYS: StorageKeys = {
```
aaronbarnardsound, line 3: Nice!

src/modules/index.ts
```diff
@@ -12,24 +12,32 @@ const defaultWalletExplanation = `Wallets are used to send, receive, and store d
...
     isMobile
   )
 
+  const {
+    heading = defaultHeading,
+    description = defaultDescription,
+    explanation = defaultWalletExplanation,
+    termsOfServiceUrl = '',
+    privacyPolicyUrl = ''
+  } = walletSelectOptions || {}
```
aaronbarnardsound, line 31: Clean! 👌

src/views/WalletSelect.svelte
```diff
@@ -29,22 +29,42 @@
...
+    description,
+    explanation,
+    termsOfServiceUrl,
+    privacyPolicyUrl,
+    wallets
+  } = module
+
+  let showTermsOfService: boolean = !!(termsOfServiceUrl || privacyPolicyUrl) && !get(app).termsAgreed
+
+  app.subscribe(({ termsAgreed }) => {
```
aaronbarnardsound, line 61:
```suggestion
const unsubscribe = app.subscribe(({ termsAgreed }) => {
```

I think we will need the unsubscribe method to unsub when this component is destroyed to avoid a memory leak

src/views/WalletSelect.svelte
```diff
@@ -29,22 +29,42 @@
...
+    if (termsAgreed) {
+      localStorage.setItem(STORAGE_KEYS.TERMS_AGREED, 'true')
+    } else {
+      localStorage.removeItem(STORAGE_KEYS.TERMS_AGREED)
+    }
+  })
```
aaronbarnardsound, line 67:
```suggestion
  })
  
  onDestroy(unsubscribe)
```

And then call it when component is destroyed

aaronbarnardsound (changes requested):
Looks good, just need to add that unsubscribe code in there I think.

taylorjdawson:
@aaronbarnardsound I ended up just refactoring to use the svelte autosubscribe/unsubscribe 👍

aaronbarnardsound (approved):

PR #551, opened 2021-05-12
1.25.0-0.1.0: [feature] gnosis safe wallet

taylorjdawson:


src/modules/select/wallets/gnosis.ts
```diff
@@ -0,0 +1,54 @@
...
+      const safe: SafeInfo | undefined = await getSafe(sdk)
+      // Checks if we are within the safe app context
+      if (!safe) return { provider: undefined, interface: null }
+
+      const provider = new SafeAppProvider(safe, sdk)
+      return {
+        provider,
+        interface: {
+          ...createModernProviderInterface(provider),
+          connect: () => Promise.resolve([safe.safeAddress])
```
taylorjdawson, line 37: The SafeAppProvider doesn't implement the `eth_requestAccounts` rpc method so this is added to mimic it in the event that it gets called

aaronbarnardsound (approved):

PR #576, opened 2021-06-07
1.31.0-0.0.3: add tp wallet

yudetamago:
Hi, I'd like to add tokenpocket (named `tp`)  wallet.
How to check if it works?

I found [react-demo](https://github.com/blocknative/react-demo) repo and added `tp` to it ( https://github.com/blocknative/react-demo/pull/66/files ) but the wallet is not displayed...

Our website is https://tokenpocket.jp/en/

Thanks!


taylorjdawson:
@yudetamago Thanks for the PR! Will take a look 👀

src/utilities.ts
```diff
@@ -285,6 +285,10 @@ export function getProviderName(provider: any): string | undefined {
     return 'Bitpie'
   }
 
+  if (provider.IsTp) {
```
taylorjdawson, line 288: @yudetamago I believe this may be your issue for the wallet not displaying in the react-demo? I think it should be lowercase: `isTp`

taylorjdawson (changes requested):

taylorjdawson:
@yudetamago When you get that change can you fix and sync with uptream branch then we can get it merged, thanks!

yudetamago:
@taylorjdawson 
Sorry for late reply, I fixed it.
Could you review this?

taylorjdawson (approved):

PR #583, opened 2021-06-11
1.28.0-0.2.0: Generic web3 wallet for injected wallet providers

mholtzman:
This pull request adds support for injected EIP-1193 providers. When a users launches a dapp with an injected provider that isn't already included as one of the provider options, an "injected" option will appear so that the user can select their custom injected provider.

image


taylorjdawson:
@mholtzman Thanks for this! We'll take a look this week. 👍🏾

src/modules/select/index.ts
```diff
@@ -65,33 +73,21 @@ function select(
...
-    return modules
-  })
+    defaultWalletNames
+      .map(walletName =>
+        walletName === 'detectedwallet' && !injectedWalletDetected()
+          ? null
+          : getModule(walletName).then((m: any) => m.default({ networkId }))
+      )
+      .filter(wallet => !!wallet)
+  )
```
mholtzman: @taylorjdawson thanks for the updates. in this block and the one above you can do the filtering first so that we don't need the awkward null filtering:

```
return Promise.all(
    defaultWalletNames
      // only include a detected wallet if it's not already one of the provided options
      .filter(walletName => walletName !== 'detectedwallet' || injectedWalletDetected())
      .map(walletName => getModule(walletName).then((m: any) => m.default({ networkId }))
  )
```

taylorjdawson: Good suggestion 👍🏾

jeremyHD:
Hi team, was wondering when will this feature be merged and release? :)

aaronbarnardsound:
@jeremyHD Just need to do some further testing on our end. If all looks good we are aiming for a release later today :)

aaronbarnardsound (approved):

PR #584, opened 2021-06-15
1.27.0-0.2.2: Refactor network check flow

### LINKED ISSUE #442: Switch wallet onboarding flow
When `walletCheck` is called and the user gets to the `network` modal, the user can click the "Switch wallet" button which will resolve the `walletCheck` call with `false` and then the wallet select modal will be displayed. In this case, there is no way for the app to know if the wallet select modal is being displayed or not. All the app knows is that the wallet check sequence was not completed.

Ideally what would happen here is that when the user clicks "Switch wallet", rather than the `walletCheck` promise resolving with `false`, it continues to run, calls `walletSelect` and waits for that to resolve. If that call resolves with `true` then it can call `walletCheck` again and proxy the result of that call to the original call to `walletCheck`. This way the app gets a complete onboarding flow and will know at all times whether there is a modal displayed or not.

taylorjdawson:
Fixes #442


src/components/Modal.svelte
```diff
@@ -2,7 +2,7 @@
   import { fade } from 'svelte/transition'
   import { app } from '../stores'
   import Branding from '../elements/Branding.svelte'
-  export let closeModal: () => void
+  export let closeModal: () => void = () => {}
```
taylorjdawson, line 5: Fixes TS error

src/elements/Spinner.svelte
```diff
@@ -1,5 +1,5 @@
-<script>
-  export let description
+<script lang="ts">
+  export let description: string = ''
```
taylorjdawson, line 2: Fixes TS error

src/views/WalletCheck.svelte
```diff
@@ -246,7 +250,7 @@
             loadingModal = false
             completed = true
             modal = res
-            resolve()
+            resolve(undefined)
```
taylorjdawson, line 253: Fixes TS error

src/views/WalletCheck.svelte
```diff
@@ -320,7 +324,7 @@
 
 {#if activeModal}
   <Modal closeModal={() => handleExit()}>
-    <ModalHeader icon={activeModal.icon} heading={activeModal.heading} />
+    <ModalHeader icon={activeModal.icon || ''} heading={activeModal.heading} />
```
taylorjdawson, line 327: Fixes TS error

aaronbarnardsound (approved):

PR #590, opened 2021-06-22
Fix derivation path validation

k06a:
Tried to make possible to use multi-digit numbers in derivation path: `m/44'/60'/12'/0/0`


aaronbarnardsound (approved):
Thanks for this :)

Issue #643, opened 2021-08-02
Fortmatic support with Polygon (matic)


ubinatus:
Hey! First of all, this library is awesome! Grats! 🥳

Back to the issue, I've noticed something strange when connecting onboard with Fortmatic on Polygon (Matic) network.

### Long story short, **Fortmatic auth iframe doesn't appear on Polygon (matic)**.

When it tries to open the Fortmatic iframe, there are some errors network requests on the console:

POST http://localhost:8545/ net::ERR_CONNECTION_REFUSED
POST https://api.rollbar.com/api/1/item/ 429. Which response is the following:
`{
  "err": 1,
  "message": "Over rate limit. Access token 68780***********80f45."
}`

_PD: No idea where that access token comes from. Not mine._ 
_PD2: I covered the token with "*" for the sake of this issue._
_PD3: Why is it requesting a local connection?._

### So, whats my onboard config?
Well, the init on the `walletSelect` for the FortMatic is the same as the demo:
`{ walletName: "fortmatic", apiKey: FORTMATIC_KEY, preferred: true }` (where FORTMATIC_KEY is the prod key from my Fortmatic dashboard)

The only difference is that I'm connecting it with a custom network, not officially supported one:
`networkId: 137`
`networkName: 'matic' `

All other wallets assigned on the `walletSelect` works like a charm. Have anyone experimented this? Or it might be something from my side?

PR #644, opened 2021-08-02
customNode on fortmatic wallet (polygon support)

ubinatus:
Following the Fortmatic guide, could be awesome to pass along a customNode on the walletSelect config for the "formatic" wallet. The purpose of this is to handle other networks connections not officially supported by blocknative/onboard (e.g. Polygon (Matic)).

The idea was taken from https://docs.fortmatic.com/web3-integration/network-configuration#switch-network-to-custom-node.


ubinatus:
Could be one fix for #643.

taylorjdawson:
Thank you @ubinatus! I opted for a slightly different approach. See #655.

ubinatus:
@taylorjdawson Awesome! Thank you

PR #648, opened 2021-08-05
1.32.0-0.2.1: [fix] Ledger Transport boolean expression assignment, log errors

mikheevm:
### Description
Ledger wallet module will ignore passed LedgerTransport in environments where WebUSB is supported, such as few latest electron versions. It is required to use a different transport there because WebUSB support is not implemented properly. I also added a few `console.error` statements for a better debugging experience

### Checklist


src/stores.ts
```diff
@@ -229,7 +229,7 @@ function createWalletStateSliceStore(options: {
   intervalSetting?: number
 }): WalletStateSliceStore {
   const { parameter, initialState, intervalSetting } = options
-  const { subscribe, set } = writable(initialState)
+  const { subscribe, set } = writable<string | number | null | undefined>(initialState)
```
mikheevm, line 232: I had a typescript error because set function from here wanted to accept only `string | number | null` as an argument if I don't explicitly set undefined

.prettierrc.js
```diff
@@ -5,6 +5,5 @@ module.exports = {
   printWidth: 80,
   tabWidth: 2,
   arrowParens: 'avoid',
-  svelteSortOrder: 'options-scripts-styles-markup',
-  plugins: ['prettier-plugin-svelte']
```
mikheevm, line 9: from plugin docs it seems it should work out of the box without adding it to the config, otherwise it didn't run the formatting for me

taylorjdawson (approved):

PR #665, opened 2021-08-18
1.34.1-0.1.1: [update] WalletConnect v1.6.2

pedrouid:
This includes fixes for a few patches which is not urgent as the 1.6.0 upgrade but useful to include on onboard.js for developers as part of the next release cycle

**1.6.1**
- upgrade ws version
- fix mobile link filtering
- fix mobile link state update
- fix react-native connector prototype

**1.6.2**
- add direct link when a single desktop link is available
- fix re-subscription on WebSocket error
- fix deep link triggered on unsafeSend method

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bnc-onboard",
-  "version": "1.34.1-0.1.0",
+  "version": "1.34.1-0.1.1",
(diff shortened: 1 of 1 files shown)


taylorjdawson:
@pedrouid Thanks! Will take a look 👀

taylorjdawson (approved):

PR #743, opened 2021-11-12
Generic provider is injected always when using a custom list

dasanra:
### Description
The generic provider is injected always when Metamask is added in the custom wallet list and also detected by onboard.js

I used the same approach that is already in the `if/else if` for the `defaultWalletNames` so they are using a similar code style

https://github.com/blocknative/onboard/blob/4751b5e3fad237b6933b8f2f356a8c0f201b4510/src/modules/select/index.ts#L124-L129

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bnc-onboard",
-  "version": "1.35.4",
+  "version": "1.35.4-0.0.1",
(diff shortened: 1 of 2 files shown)


taylorjdawson (approved):
@dasanra Excellent, thanks! 🙏🏾

taylorjdawson:
Fixes #743

PR #755, opened 2021-12-03
Fix/circleci npm publish

taylorjdawson:
Fixes the path issue


Adamj1232 (approved):
Nice man! Good catch!!

PR #763, opened 2021-12-20
WIP: Feature/ledger wallet

taylorjdawson:


packages/demo/webpack.config.js
```diff
@@ -57,6 +57,9 @@ module.exports = {
       }
     ]
   },
+  experiments: {
+    asyncWebAssembly: true,
```
taylorjdawson, line 61: Needed for npm module: `tiny-secp256k1`

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,270 @@
...
+}
+
+const getAddress = async (
+  { publicKey, chainCode, derivationPath }: LedgerAccount,
+  asset: Asset,
+  { rpcUrl }: Chain,
+  index: number
+): Promise<Account> => {
+  const { BIP32Factory } = await import('bip32')
+  const ecc = await import('tiny-secp256k1')
```
taylorjdawson, line 78: This is used in place of `hdkey` which is outdated and isn't written in Typescript

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,270 @@
...
+        const { rlp } = (await import('ethereumjs-util')).default
+
+        const transport: Transport = await getTransport()
+        const eth = new Eth(transport)
+        const eventEmitter = new EventEmitter()
+
+        let currentChain: Chain | undefined
+        const scanAccounts = async ({
+          derivationPath,
+          chainId,
```
taylorjdawson, line 140: @aaronbarnardsound As far as I can tell this is the only way we know what chainId the user has selected correct?

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,270 @@
...
+}
+
+const getBalance = async (
+  address: string,
+  rpcUrl: string,
+  block: string | number = 'latest'
+): Promise<string> => {
+  const { providers } = await import('ethers')
+  const provider = new providers.JsonRpcProvider(rpcUrl)
+  return (await provider.getBalance(address, block)).toHexString()
```
taylorjdawson, line 68: @aaronbarnardsound Balance in hex or string number? TS Types currently are string. IMO it should stay bignumber until it gets displayed so that it stays flexible up until it is time to be displayed

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        const getAccounts = async () =>
+          accounts ??
+          (accounts = await accountSelect({
+            basePaths: DEFAULT_BASE_PATHS,
+            assets,
+            chains,
+            scanAccounts,
+            walletIcon: '<svg></svg>'
```
taylorjdawson, line 173: @aaronbarnardsound Wasn't sure how this is different from the `getIcon` method above? 🤔

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,270 @@
...
+            // Triggers the account select modal if no accounts have been selected
+            const [{ address }] = await getAccounts()
+            return [address]
+          },
+          eth_accounts: async baseRequest => {
+            return accounts?.[0]?.address ? [accounts[0].address] : []
+          },
+          eth_chainId: async baseRequest => {
+            console.log('eth_chainId called', currentChain?.id)
+            return currentChain?.id ?? ''
```
taylorjdawson, line 188: @aaronbarnardsound I am noticing this method get's called before the `scanAccounts` method so the chainId isn't set yet -- is this desired behaviour?

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,261 @@
...
+              )
+              let unsignedTx = transaction.getMessageToSign(false)
+
+              // If this is not an EIP1559 transaction then it is legacy and it needs to be
+              // rlp encoded before being passed to ledger
+              if (!transaction.supports(Capability.EIP1559FeeMarket)) {
+                unsignedTx = rlp.encode(unsignedTx)
+              }
+
+              const [{ derivationPath }] = accounts || [{ derivationPath: '' }]
```
taylorjdawson, line 218: @aaronbarnardsound The JSON RPC API for eth_signTransaction has a `from` address. Do we just ignore this field and use the user selected account address as the `from` field? 
In current onboard it looks like we don't touch the `from` field on the transaction object and assume that the `from` field is the same as the first path. ([source](https://github.com/blocknative/onboard/blob/e3f04c43ed705c56099a91e5d2be704fb38599e8/src/modules/select/wallets/ledger.ts#L407))

packages/ledger/src/index.ts
aaronbarnardsound: Yeah essentially the accountSelect module needs the wallet icon to render it in the modal. If you could call the `getIcon` method and then pass the icon in here that would be great.

packages/ledger/src/index.ts
aaronbarnardsound: This should be fixed in the latest commits

packages/ledger/src/index.ts
aaronbarnardsound: I think if there is a from field, we map that to the derivation path and use that, if not just use the first path

aaronbarnardsound (changes requested):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+  }
+]
+
+const assets = [
+  {
+    label: 'ETH'
+  }
+]
+
+const getFullDerivationPath = (derivationPath: string, index: number): string =>
```
aaronbarnardsound, line 31: This seems to be unused, so I think we could remove?

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,339 @@
+import { accountSelect, createEIP1193Provider } from '@bn-onboard/common'
```
aaronbarnardsound, line 1: Could we dynamically import these inside the `getInterface` function?

packages/ledger/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+            }))
+          )
+        }
+
+        const ledgerProvider = {}
+
+        const provider = createEIP1193Provider(ledgerProvider, {
+          eth_requestAccounts: async baseRequest => {
+            // Triggers the account select modal if no accounts have been selected
+            const [{ address }] = await getAccounts()
```
aaronbarnardsound, line 187: If the user exits the account select modal without selecting an account, then the `getAccounts` function will return an empty array. I think we should add a check to make sure the result has `length`, and if not then we need to throw the error code for when the user rejects connection to be compliant with the `eth_requestAccounts` spec. In the error case, we will also need to see the `account` variable to undefined to indicate that accounts have not yet been connected.

aaronbarnardsound (changes requested):
A few comments and some requested changes :)

aaronbarnardsound (approved):

aaronbarnardsound (approved):

Pojielp (approved):

Pojielp (approved):

Pojielp (approved):

PR #776, opened 2022-01-12
1.37.0-0.0.1 : [Update] - Walletlink version to 2.4.2

brendanww:
### Description
Updates walletlink to a version of the protocol that supports returning a 4092 error code for `switchEthereumChain` requests where the network is not recognized by the wallet. This enables dapps to send an `addEthereumChain` request after receiving the 4092 error code.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bnc-onboard",
-  "version": "1.37.0",
+  "version": "1.37.0-0.0.1",
   "description": "Onboard users to web3 by allowing them to select a wallet, get that wallet ready to transact and have access to synced wallet state.",
   "keywords": [
     "ethereum",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bnc-onboard",
-  "version": "1.37.0",
+  "version": "1.37.1",
```
Adamj1232, line 3: @brendanww For versioning going into develop we add a development version, in this case would look like 1.37.0-0.0.1 and inc version after testing in stage when publishing/releasing
If you could make that update that would be awesome. Thanks man!!

Adamj1232 (changes requested):

package.json
brendanww: done! thanks for the quick feedback

Adamj1232 (approved):
Thank you! Merging to develop now and will get into master shortly

Adamj1232 (changes requested):
@brendanww It looks like your yarn lock changes have removed the integrity hash.
Do you happen to have an older version of yarn or a setting that doesn't populate that field?
We currently run yarn version 1.22.17
Once fixed we can get this merged and deployed quickly, thank you!

brendanww:
@Adamj1232 Updated! thank you

Adamj1232 (approved):

PR #792, opened 2022-01-22
feature: auto add RPC if not already added in MetaMask

jadbox:
### Description
This PR tries to auto add Polygon as a network to Metamask if it's not already configured in the users wallet.

See this issue for more context: https://github.com/blocknative/onboard/issues/791

### Checklist
- [ ] This PR passes the Circle CI checks (if **merging a fork** run `yarn` with node version 12.22.7 to ensure no errors)


taylorjdawson:
@jadbox Shameless plug 🔌  but this would be a really good use case for [eth-chains](https://github.com/taylorjdawson/eth-chains) so you don't need to hardcode all of those networks

jadbox:
@taylorjdawson I was thinking about adding am RPC library like this, but I wanted to keep it simple for now to reduce possible bugs and also to avoid adding another dependency at his stage. One issue that I spotted with using `eth-chains` is several of the 1st rpc providers have a template variable (`'https://mainnet.infura.io/v3/${INFURA_API_KEY}`) that would require further work to have a resolver for before passing it to MetaMask.

jadbox:
btw, I've confirmed the code works to prompt the user to add the network. I've published my fork to: https://www.npmjs.com/package/@mintgate/bnc-onboard

Prompt looks like this:
Screenshot 2022-01-22 114325

nosyajmagS020796:
Pls return my payment

PR #796, opened 2022-01-24
[feature] Keystone Module

taylorjdawson:


packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+                message: 'User rejected the request.'
+              })
+            }
+            return [accounts[0]?.address]
+          },
+          eth_accounts: async () => {
+            return accounts?.[0]?.address ? [accounts[0].address] : []
+          },
+          eth_chainId: async () => {
+            return currentChain?.id ?? ''
```
aaronbarnardsound, line 143:
```suggestion
            return currentChain.id
```

O2 core enforces that at least one chain must be set before calling `connectWallet`, so you can be sure that it exists.

packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+          },
+          eth_signTransaction: async ({ params: [transactionObject] }) => {
+            if (!accounts)
+              throw new Error(
+                'No account selected. Must call eth_requestAccounts first.'
+              )
+
+            const account =
+              accounts.find(
+                account => account.address === transactionObject?.from
```
aaronbarnardsound, line 153:
```suggestion
                account => account.address === transactionObject.from
```
You could also add a check that `transactionObject` exists at the top and throw an error if it does not as is required

packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+              ) || accounts[0]
+
+            const { address: from } = account
+
+            // Set the `from` field to the currently selected account
+            transactionObject = { ...transactionObject, from }
+
+            const common = new Common({
+              chain: customNetwork || Number.parseInt(currentChain?.id) || 1,
```
aaronbarnardsound, line 162:
```suggestion
              chain: customNetwork || Number.parseInt(currentChain.id) || 1,
```

packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+              // Berlin is the minimum hardfork that will allow for EIP1559
+              hardfork: Hardfork.Berlin,
+              // List of supported EIPS
+              eips: [1559]
+            })
+
+            transactionObject.gasLimit =
+              transactionObject.gas ?? transactionObject.gasLimit
```
aaronbarnardsound, line 170:
```suggestion
              transactionObject.gas || transactionObject.gasLimit
```

`gas` is either a string or is `undefined`

packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+              {
+                ...transactionObject
+              },
+              { common, freeze: false }
+            )
+
+            // @ts-ignore
+            const signedTx = await keyring.signTransaction(from, transaction)
+
+            return `0x${signedTx?.serialize().toString('hex')}`
```
aaronbarnardsound, line 182:
```suggestion
            return `0x${signedTx.serialize().toString('hex')}`
```
If `signedTx` is `undefined` here we have problems regardless and should handle it differently

packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+          },
+          eth_sign: async ({ params: [address, message] }) => {
+            if (!(accounts?.length && accounts?.length > 0))
```
aaronbarnardsound, line 185:
```suggestion
            if (!(accounts.length && accounts.length > 0))
```

packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+              )
+
+            const account =
+              accounts.find(account => account.address === address) ||
+              accounts[0]
+
+            return keyring.signMessage(account.address, message)
+          },
+          eth_signTypedData: async ({ params: [address, typedData] }) => {
+            if (!(accounts?.length && accounts?.length > 0))
```
aaronbarnardsound, line 197:
```suggestion
            if (!(accounts.length && accounts.length > 0))
```

packages/keystone/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+
+            const account =
+              accounts.find(account => account.address === address) ||
+              accounts[0]
+
+            return keyring.signTypedData(account.address, typedData)
+          },
+          wallet_switchEthereumChain: async ({ params: [{ chainId }] }) => {
+            currentChain =
+              chains.find(({ id }) => id === chainId) ?? currentChain
```
aaronbarnardsound, line 210:
```suggestion
              chains.find(({ id }) => id === chainId) || currentChain
```

aaronbarnardsound (changes requested):
Will need to test once the requested changes are made, but looks good so far.

aaronbarnardsound (approved):

PR #841, opened 2022-02-25
1.38.1-0.0.3: [fix] Detected wallet logic

### LINKED ISSUE #765: [Bug]: Detected wallet option no longer appears in select modal
### Current Behavior

We use a custom Chrome extension that injects an EIP-1193 provider in `window.ethereum`. Up until recently, it was showing up as "Detected" option in the select modal when browsing Dapps using onboard (like curve.fi). This stopped working with version `1.35.5`, I believe.

If I understand https://github.com/blocknative/onboard/pull/743 correctly, the `detectedwallet` option must now be explicitly enabled to show up. This might be deliberate, but it's still a breaking change.

### Expected Behavior

A generic injected providers should continue showing up as "Detected" option in the select modal as there are users who came to rely on this.

The "Detected" option should not show up if the injected provider is identified as one of the known ones like MetaMask, effectively fixing the bug https://github.com/blocknative/onboard/issues/740 without introducing a breaking change.

### Steps To Reproduce

_No response_

### Onboard Version

^1.35.5

### Node Version

14.17.0

Chrome

### Relevant log output

_No response_

### Anything else?

_No response_

taylorjdawson:
### Description
Logic needed to be updated to check that the provider exists rather than if the name of the provider exists. That defeated the purpose of a "detected wallet".

Fixes #765

### Checklist
- [x] This PR passes the Circle CI checks (if **merging a fork** run `yarn` with node version 12.22.7 to ensure no errors)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bnc-onboard",
-  "version": "1.38.1-0.0.2",
+  "version": "1.38.1-0.0.3",
   "description": "Onboard users to web3 by allowing them to select a wallet, get that wallet ready to transact and have access to synced wallet state.",
   "keywords": [
     "ethereum",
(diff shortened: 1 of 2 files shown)


Adamj1232 (approved):
LGTM!! 🚀 BlazeIt 🔥

Issue #857, opened 2022-03-04
[Bug]: Ens not populating avatar


sherodtaylor:
### Current Behavior

- ENS object on `Accounts` is not populating avatar for  Mainnet Ethereum 
- ENS object on `Accounts` is missing all fields on Rinkbey chain

### Onboard Version

2.0.2

### Node Version

v16.13.2

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell
Mainnet

{
    "name": "sherodbuilds.eth",
    "contentHash": null
}
```
Rinkbey ens empty
```
null
```
```


### Anything else?

_No response_

aaronbarnardsound:
Ah yes I can see that we have missed `avatar` on the ENS object, I will get that fixed in the next release very soon. 

For the Rinkeby ENS, does it show up after a few moments? The state gets set to null initially and then gets updated once the network requests resolve.

aaronbarnardsound:
Closed via #862

sherodtaylor:
Thank you!

PR #862, opened 2022-03-07
core-v2.0.6: [fix] - Add Avatar to ENS Object

aaronbarnardsound:
### Description
- Adds missing `avatar` property to ENS object


aaronbarnardsound:
Fixes #857

PR #873, opened 2022-03-09
web3-onboard/core:v2.0.9: [enhancement] - Autoselect Disable Modals

### LINKED ISSUE #849: [Feature]: calling onboard.connectWallet with the autoSelect option should not flash the modal
### Is your request related to a problem?

_No response_

### Feature Description

For background context I am following the example provided here: https://docs.blocknative.com/onboard/core#auto-selecting-a-wallet

Auto-selecting is working well, however when the page refreshes, or the user navigates to a different route - the modal flashes for a second, immediately displays a 'Connected' message, and then closes. 

Ideally the library would allow controlling this behavior so that page refresh and navigation could occur without the modal opening/closing - which would make the app feel a lot smoother from a UX perspective. 

I have attached a video depicting this behavior below. 

Thanks in advance and let me know if there's anything else I can provide! 

### Alternative Solutions

_No response_

### Anything else?

Video of page refresh and navigation: 
https://user-images.githubusercontent.com/28926919/156272168-1a1eae32-6ebb-4025-88a0-d9ad2a174850.mp4

aaronbarnardsound:
### Description
- Lifts all logic out of sub modal components in to the main index for the connect flow
- Modifies type and validation to allow for an options object for the `autoselect` parameter so that modals can be disabled when auto selecting a wallet:

```javascript
const previouslySelectedWallet = window.localStorage.getItem('selectedWalletLabel')

onboard.connectWallet({
  autoSelect: {
    label: previouslyConnectedWallet,
    disableModals: true
  }
})
```

The original way of calling connect wallet with the `autoselect` parameter as a string is also still valid and will get converted to the new options object internally.

Closes #849

### DIFF
--- a/packages/core/src/views/connect/Index.svelte
+++ b/packages/core/src/views/connect/Index.svelte
@@ -6,14 +6,17 @@
   import en from '../../i18n/en.json'
   import { selectAccounts } from '../../provider'
   import { state } from '../../store'
-  import { addWallet } from '../../store/actions'
   import { connectWallet$, internalState$ } from '../../streams'
-  import type {
-    i18n,
-    WalletState,
-    WalletWithLoadedIcon,
-    WalletWithLoadingIcon
-  } from '../../types'
+  import {
+    getChainId,
+    requestAccounts,
+    trackWallet,
+    getBalance,
+    getEns
+  } from '../../provider'
+  import { addWallet, updateAccount } from '../../store/actions'
+  import { validEnsChain } from '../../utils'
+
   import CloseButton from '../shared/CloseButton.svelte'
   import Modal from '../shared/Modal.svelte'
   import Agreement from './Agreement.svelte'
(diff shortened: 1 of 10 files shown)


packages/core/src/views/connect/Index.svelte
```diff
@@ -22,87 +25,103 @@
...
+  export let autoSelect: ConnectOptions['autoSelect']
 
   const { walletModules, appMetadata } = internalState$.getValue()
 
-  let loading = true
   let connectionRejected = false
   let wallets: WalletWithLoadingIcon[] = []
   let selectedWallet: WalletState | null
   let agreed: boolean
+  let connectingWallet: string // the wallet label that is connecting
```
Adamj1232, line 43: you could add `label` to the variable and then the comments wouldnt be needed

Adamj1232 (approved):
Love this change!!
Small comment but not necessary for merge

packages/core/src/views/connect/Index.svelte
aaronbarnardsound: Yeah good call

PR #879, opened 2022-03-10
Magic wallet module addition

Adamj1232:
### Description
Magic wallet module addition

### Checklist


packages/magic/src/index.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+        try {
+          loggedIn = await magicInstance.user.isLoggedIn()
+        } catch (err) {
+          throw new Error(`$An error occured while connecting your Magic wallet, please try again: ${err}`)
+        }
+        
+        if (!loggedIn) handleLogin()
+        
+        let magicProvider = new Web3(magicInstance.rpcProvider)
```
aaronbarnardsound, line 61: I don't think Web3 is needed here. Can we just use the `magicInstance.rpcProvider` and pass that to be patched?

packages/magic/src/index.ts
```diff
@@ -0,0 +1,167 @@
...
+
+          return provider
+        }
+
+        provider = patchProvider()
+
+        return {
+          provider,
+          magicInstance
+        }
```
aaronbarnardsound, line 161:
```suggestion
        return {
          provider,
          instance: magicInstance
        }
```

aaronbarnardsound (changes requested):
Looking good, just a couple of small changes!

packages/magic/src/index.ts
Adamj1232: Done!

packages/magic/src/index.ts
Adamj1232: Done and done!

packages/demo/src/App.svelte
```diff
@@ -169,7 +175,7 @@
   const signTypedMessage = async (provider, address) => {
     const data = JSON.parse(signTypedMsg)
     const signature = await provider.request({
-      method: 'eth_signTypedData',
+      method: 'eth_signTypedData_v4',
```
aaronbarnardsound, line 178:
```suggestion
      method: 'eth_signTypedData',
```

packages/magic/README.md
```diff
@@ -0,0 +1,41 @@
...
+  // ... other Onboard options
+  wallets: [
+    magic
+    //... other wallets
+  ]
+})
+
+const connectedWallets = await onboard.connectWallet()
+console.log(connectedWallets)
+```
```
aaronbarnardsound, line 41: Need some documentation in here that explains how to get access to the user profile data from the instance.

packages/magic/rollup.config.js
```diff
@@ -0,0 +1,39 @@
...
+    resolve({
+      browser: true,
+      dedupe: ['svelte']
+    }),
+    typescript({
+      sourceMap: !production,
+      inlineSources: !production
+    })
+  ],
+  external: ['rxjs', 'web3', 'magic-sdk', '@web3-onboard/common']
```
aaronbarnardsound, line 38:
```suggestion
  external: ['rxjs', 'magic-sdk', '@web3-onboard/common']
```

packages/magic/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+          ProviderRpcErrorCode,
+          ProviderRpcError
+        } = await import('@web3-onboard/common')
+
+        const emitter = new EventEmitter()
+
+        let currentChain = chains[0].id
+
+        let customNodeOptions = {
+          chainId: parseInt(chains[0].id, 10),
```
aaronbarnardsound, line 28:
```suggestion
          chainId: parseInt(chains[0].id),
```
I had to fix this recently in other modules where I was using the radix incorrectly which causes the number to be parsed incorrectly.

packages/magic/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+
+        const handleLogin = async () => {
+          loggedIn = await loginModal({
+            walletName: walletName,
+            brandingHTMLString: brandingHTML,
+            emailLoginFunction: loginWithEmail
+          })
+        }
+
+        if (!loggedIn) await handleLogin()
```
aaronbarnardsound, line 55: Does this call to login need to be here? The `core` module will call the `eth_requestAccounts` once it gets the provider, so i think that `handleLogin` might be called twice in quick succession.

packages/magic/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+                return []
+              }
+            },
+            eth_selectAccounts: null,
+            eth_getBalance: async () => {
+              const address = (
+                await magicProvider.request({
+                  method: 'eth_accounts'
+                })
+              )[0]
```
aaronbarnardsound, line 87:
```suggestion
              const [address] = (
                await magicProvider.request({
                  method: 'eth_accounts'
                })
              )
```
😉

packages/magic/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+              const chain = chains.find(({ id }) => id === params[0].chainId)
+              if (!chain) throw new Error('Chain must be set before switching')
+              currentChain = chain.id
+              // re-instantiate instance with new network
+              customNodeOptions = {
+                chainId: parseInt(chains[0].id, 10),
+                rpcUrl: chains[0].rpcUrl
+              }
+
+              magicInstance = new Magic(apiKey, {
```
aaronbarnardsound, line 106: Did you get a chance to check if we need to do any cleanup on the original instance and or provider to prevent any possible memory leaks?

packages/magic/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+            eth_accounts: async () => {
+              const accounts = (
+                await magicProvider.request({
+                  method: 'eth_accounts'
+                })
+              )
+              return Array.isArray(accounts) && accounts.length
+                ? [accounts[0]]
+                : []
+            },
```
aaronbarnardsound, line 127:
```suggestion
```

I believe that we can just remove this and not patch it as it seems like the method is to spec? Is there a situation where it does not return an array?

packages/magic/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+            eth_chainId: async () => {
+              return currentChain ? currentChain : ''
+            },
+
+            eth_signTransaction: async ({ params: [transactionObject] }) => {
+              const fromAddress = (
+                await magicProvider.request({
+                  method: 'eth_accounts'
+                })
+              )[0]
```
aaronbarnardsound, line 159:
```suggestion
              const [fromAddress] = (
                await magicProvider.request({
                  method: 'eth_accounts'
                })
              )
```

aaronbarnardsound (changes requested):

packages/magic/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        const {
+          createEIP1193Provider,
+          ProviderRpcErrorCode,
+          ProviderRpcError
+        } = await import('@web3-onboard/common')
+
+        const emitter = new EventEmitter()
+
+        let currentChain = chains[0].id
```
taylorjdawson, line 25: Is there scenario where `chains` is empty?

packages/magic/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+          network: customNodeOptions
+        })
+
+        const loginWithEmail = async (emailAddress: string) => {
+          try {
+            await magicInstance.auth.loginWithMagicLink({ email: emailAddress })
+            return await magicInstance.user.isLoggedIn()
+          } catch (err) {
+            throw new Error(
+              `$An error occurred while connecting your Magic wallet, please try again: ${err}`
```
taylorjdawson, line 42: Is the `$` sign here in the error message on purpose?

packages/magic/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+          provider = createEIP1193Provider(magicProvider, {
+            eth_requestAccounts: async ({ baseRequest }) => {
+              try {
+                if (!loggedIn) await handleLogin()
+                const accounts = await baseRequest({ method: 'eth_accounts' })
+                return accounts
+              } catch (error) {
+                console.error('error in request accounts', error)
+                const { code } = error as { code: number }
+                if (code === -32603) {
```
taylorjdawson, line 68: Could we use an enum or constant here?

packages/magic/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+              const receipt = await magicProvider.send('eth_sign', [
+                address,
+                message
+              ])
+              return receipt &&
+                receipt.hasOwnProperty('tx') &&
+                receipt.tx.hasOwnProperty('hash')
+                ? receipt.tx.hash
+                : ''
+            },
```
taylorjdawson, line 130: Might be able to shorten this method!?

```suggestion
            eth_sign: async ({ params }) =>  
              magicProvider.send('eth_sign', params).then( receipt => 
                receipt && receipt.hasOwnProperty('tx') && 
                receipt.tx.hasOwnProperty('hash') ? receipt.tx.hash : ''
              )
```

packages/magic/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+            eth_signTypedData: async ({ params: [address, typedData] }) => {
```
taylorjdawson, line 131: @Adamj1232 same as above, maybe this can be shortened and you can leave out the array&object destructuring -- seems like an unnecessary step

packages/magic/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+              ])
+              return receipt &&
+                receipt.hasOwnProperty('tx') &&
+                receipt.tx.hasOwnProperty('hash')
+                ? receipt.tx.hash
+                : ''
+            },
+            eth_chainId: async () => {
+              return currentChain ? currentChain : '0x1'
+            },
```
taylorjdawson, line 144: Maybe?
```suggestion
            eth_chainId: async () => currentChain || '0x1',
```

taylorjdawson (changes requested):
@Adamj1232 I added a few suggestions but nothing too critical!

packages/demo/src/App.svelte
```diff
@@ -79,8 +80,13 @@
   }
   const trezor = trezorModule(trezorOptions)
 
+  const magic = magicModule({
+    apiKey: 'pk_live_02207D744E81C2BA'
```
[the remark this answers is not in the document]
Adamj1232: Yessir, it will be removed or just left in for demoing fun

packages/magic/src/view/LoginModal.svelte
```diff
@@ -0,0 +1,199 @@
...
+  } = loginOptions
+
+  let errorInEmail = false
+
+  const setErrorInEmail = () => {
+    if (!errorInEmail) return
+    errorInEmail = false
+  }
+
+  const emailRegex =
```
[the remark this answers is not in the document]
Adamj1232: We dont have anywhere else that we check email in onboard yet but I agree it can be moved to a more general location when we need it

packages/magic/src/index.ts
Adamj1232: I removed the destructuring since the params are in the the proper order for the calls but kept the await instead of the .then
Good call on the destructuring, I was thinking the same thing but wasn't sure if that would hurt readability and since you are one of the readers you confirmed 💯

packages/magic/src/index.ts
Adamj1232: Yep you are correct, removed!

packages/common/src/types.ts
```diff
@@ -228,7 +228,8 @@ export type ChainId = string
 export type RpcUrl = string
 
 export type WalletInterface = {
-  provider: EIP1193Provider
+  provider: EIP1193Provider,
+  instance?: unknown
```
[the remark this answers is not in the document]
Adamj1232: I generally agree with this but the Magic instance that we are passing back is evolving quickly with development so keeping as unknown to allow for less breakage and updating

packages/magic/src/index.ts
Adamj1232: We should be good as is without causing any extra calls

packages/magic/README.md
Adamj1232: Good call!! Documentation added!

packages/magic/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "@tsconfig/svelte/tsconfig.json",
```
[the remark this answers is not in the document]
Adamj1232: Wanted checking with @aaronbarnardsound on this comment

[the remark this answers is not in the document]
aaronbarnardsound: I don't think we want to extend from the base since it is quite different and there may be conflicts. The `core` and `common` packages both use this svelte specific ts config.

packages/common/src/types.ts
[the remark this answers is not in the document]
aaronbarnardsound: It also needs to cover all SDK wallet instances which we cannot type all of them

packages/magic/README.md
```diff
@@ -0,0 +1,65 @@
...
+  instance
+}
+
+// Assumes a user is already logged in
+try {
+  const { email, publicAddress } = await m.user.getMetadata();
+} catch {
+  // Handle errors if required!
+}
+```
```
aaronbarnardsound, line 64: I assume the `m` in your example refers to the instance, so I think we want to provide a more concrete example like this:

```suggestion
```typescript
const [magicWallet] = await onboard.connectWallet()

try {
  const { email, publicAddress } = await magicWallet.instance.user.getMetadata();
} catch {
  // Handle errors if required!
}
```
```

File: packages/demo/src/App.svelte [hunk 23]
```diff
@@ -79,8 +80,13 @@
...
+  })
+
   const onboard = Onboard({
     wallets: [
+      magic,
```
aaronbarnardsound, line 89: I think we can place this further down the list for the demo, maybe just in front of the Fortmatic module?

packages/core/src/views/connect/Index.svelte
```diff
@@ -111,6 +111,9 @@
         chains: [{ namespace: 'evm', id: '0x1' }]
       }
 
+      if (instance)
+        selectedWallet.instance = instance
+
```
aaronbarnardsound, line 116:
```suggestion
```

We can remove this block and just add the instance to the selected wallet since it is optional in the types so it is fine if it is `undefined`:

```javascript
selectedWallet = {
  label,
  icon: loadedIcon,
  provider,
  instance,
  accounts: [],
  chains: [{ namespace: 'evm', id: '0x1' }]
}
```

packages/core/src/validation.ts
```diff
@@ -57,6 +57,7 @@ const wallet = Joi.object({
   label: Joi.string(),
   icon: Joi.string(),
   provider: unknownObject,
+  instance: unknownObject.optional(),
```
aaronbarnardsound, line 60:
```suggestion
  instance: unknownObject,
```
Joi validations default to optional so this can be removed.

packages/magic/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+        const emitter = new EventEmitter()
+
+        if (!chains.length)
+          throw new Error(
+            'Atleast one Chain must be passed to onboard in order to connect'
+          )
+
+        let currentChain = chains[0]
+        console.log(currentChain)
```
aaronbarnardsound, line 32:
```suggestion
```

Sneaky log

packages/magic/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+        function patchProvider(): EIP1193Provider {
+          const patchedProvider = createEIP1193Provider(magicProvider, {
+            eth_requestAccounts: async ({ baseRequest }) => {
+              try {
+                if (!loggedIn) await handleLogin()
+                const accounts = await baseRequest({ method: 'eth_accounts' })
+                activeAddress = accounts[0]
+                return accounts
+              } catch (error) {
+                console.error('error in request accounts', error)
```
aaronbarnardsound, line 75:
```suggestion
```
Sneaky log

packages/magic/src/login-modal.ts
```diff
@@ -0,0 +1,98 @@
...
+) => {
+  class loginModalEl extends HTMLElement {
+    constructor() {
+      super()
+    }
+  }
+
+  if (!customElements.get('login-modal')) {
+    customElements.define('login-modal', loginModalEl)
+  }
```
aaronbarnardsound, line 41:
```suggestion
  if (!customElements.get('onboard-magic-login-modal')) {
    customElements.define('onboard-magic-login-modal', loginModalEl)
  }
```
To be safe I reckon we need to have a more specific namespace for this element.

packages/magic/src/login-modal.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  styleEl.innerHTML = `
+    ${SofiaProRegular}
+    ${SofiaProSemiBold}
+    ${SofiaProLight}
+  `
+  document.body.appendChild(styleEl)
+
+  // add to DOM
+  const loginModalDomElement = document.createElement('login-modal')
```
aaronbarnardsound, line 54:
```suggestion
  const loginModalDomElement = document.createElement('onboard-magic-login-modal')
```

aaronbarnardsound (changes requested):
We need to increment the version on the `core` package since there are changes in there as well as a few minor changes indicated in the comments.

packages/magic/src/login-modal.ts
Adamj1232: donezo

aaronbarnardsound (approved):
Happy days!!

PR #886, opened 2022-03-17
fix missing initial states in react hooks

whatsyourgithub:
### Description
This sets initial react state in the hooks, instead of defaulting to null.

Right now, if a hook is used in a component that renders after you connect your wallet, it won't have any wallets in it's react state. Unless a wallet change happens afterwards that triggers the subscription to update it. Initially, it's all null.
This problem applies to all 3 react hooks.

### Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/react",
-  "version": "2.0.5",
+  "version": "2.0.4-0.0.1",
   "description": "Collection of React Hooks for web3-onboard",
   "module": "dist/index.js",
   "browser": "dist/index.js",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/react",
-  "version": "2.0.4-0.0.1",
+  "version": "2.0.5",
   "description": "Collection of React Hooks for web3-onboard",
   "module": "dist/index.js",
   "browser": "dist/index.js",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/react",
-  "version": "2.0.4",
+  "version": "2.0.5",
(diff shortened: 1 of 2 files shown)


packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/react",
-  "version": "2.0.4",
+  "version": "2.0.5",
```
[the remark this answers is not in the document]
whatsyourgithub: Fixed

whatsyourgithub:
Note that we're using these changes in a live app and all is working as expected

packages/react/src/index.ts
```diff
@@ -89,7 +83,15 @@ export const useSetChain = (
...
 
   const [connectedChain, setConnectedChain] = useState<ConnectedChain | null>(
-    null
+    () => {
+      const initialWallets = (web3Onboard as OnboardAPI).state.get().wallets
+      return (
+        (
+          initialWallets.find(({ label }) => label === walletLabel) ||
+          initialWallets[0]
+        )?.chains[0] || null
```
Adamj1232, line 92: Can you remove the optional chaining (?.) when you have a moment?
We have to stick with a bit older ES operators in order to support all of our customers bundles and compilers

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/react",
-  "version": "2.0.4",
+  "version": "2.0.4-0.0.1",
```
Adamj1232, line 3: Apologies on this but you can revert to `2.0.5` here
We have temporarily moved away from feature branch versioning while we get out staging env build flow figured out.
Sorry about the requested change and revert.

Adamj1232 (approved):
@whatsyourgithub Thank you for the updates!
We are always looking to learn more about the projects who use our products!
Feel free to hit me up on discord Adrock86#7378 or through my email adam@blocknative.com 
Cheers and thanks again!

PR #906, opened 2022-03-28
[core-v2.1.0] - Feature: Update Wallets without Re-Initialization

aaronbarnardsound:
### Description
This PR makes it possible for an app to update the wallets that are displayed in the wallet selection modal after initialization, without needing to re-initialize. The main changes that achieve this are:

- Move the `walletModules` state from the `internalState$` stream, to the main state store and update references to the new location
- Add an action to update the `walletModules` state
- Expose the action under a new `actions` parameter that lives on the state API

### Checklist


Adamj1232 (approved):
Looks good! I like this change a lot, the ability to expose this to DApps will allow for optimal UX flows.
Will we hit the ReactHook package with this version update?

PR #909, opened 2022-03-29
use tally only namespace

greg-nagy:
### Description

The infrasture is there to be able to detect
multiple injected provider which is awesome!
I am honestly grateful for this!

But if we use mostly window.ethereum we
kinda beat the purpose of the whole infrasturcure
being there.

So let's use wallet specific namespaces.

### Checklist
> This is a PR for a feature branch, I assume this does not apply here
> I think I can't do that on an organization owned fork :(
> Tested with https://github.com/blocknative/react-demo and `yarn link @web3-onboard/injected-wallets ` and the latest main on tally 
> Not sure :(

### DIFF
--- a/packages/injected/src/wallets.ts
+++ b/packages/injected/src/wallets.ts
@@ -445,19 +445,17 @@ const tokenary: InjectedWalletModule = {
 
 const tally: InjectedWalletModule = {
   label: ProviderLabel.Tally,
-  injectedNamespace: InjectedNameSpace.Ethereum,
+  injectedNamespace: InjectedNameSpace.Tally,
   checkProviderIdentity: ({ provider }) =>
     !!provider && !!provider[ProviderIdentityFlag.Tally],
   getIcon: async () => (await import('./icons/tallywallet.js')).default,
   getInterface: async () => {
-    const provider = createEIP1193Provider(window.ethereum, {
-      eth_chainId: ({ baseRequest }) => baseRequest({ method: 'eth_chainId' }).then(id => `0x${parseInt(id).toString(16)}`),
-    });
-    provider.removeListener = (event, func) => { };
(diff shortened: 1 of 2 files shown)


packages/injected/src/wallets.ts
```diff
@@ -445,19 +445,17 @@ const tokenary: InjectedWalletModule = {
...
 const tally: InjectedWalletModule = {
   label: ProviderLabel.Tally,
-  injectedNamespace: InjectedNameSpace.Ethereum,
+  injectedNamespace: InjectedNameSpace.Tally,
   checkProviderIdentity: ({ provider }) =>
     !!provider && !!provider[ProviderIdentityFlag.Tally],
   getIcon: async () => (await import('./icons/tallywallet.js')).default,
   getInterface: async () => {
-    const provider = createEIP1193Provider(window.ethereum, {
-      eth_chainId: ({ baseRequest }) => baseRequest({ method: 'eth_chainId' }).then(id => `0x${parseInt(id).toString(16)}`),
```
greg-nagy, line 454: We have landed a fix for this [in this PR](https://github.com/tallycash/extension/pull/1247) which will be also in the next release. Is currently on main.

packages/injected/src/wallets.ts
```diff
@@ -445,19 +445,17 @@ const tokenary: InjectedWalletModule = {
...
-    });
-    provider.removeListener = (event, func) => { };
+    const provider = createEIP1193Provider(window.tally)
+    provider.removeListener = (event, func) => {}
```
greg-nagy, line 454: Can you please describe why you needed this?

packages/injected/src/wallets.ts
Adamj1232: @greg-nagy excellent,  will that change be in the upcoming release? ETA on the release?

PR #917, opened 2022-04-05
Feature: Account Center UI

aaronbarnardsound:
### Description
**[core v2.1.0]**
- Adds a new UI for users to manage their connected accounts called the "Account Center"
- Refactors repeated CSS into global classes
- Replaces icons with bounding boxes

*Minimized UI:*
Screen Shot 2022-04-05 at 12 47 34 pm

*Maximized UI:*
Screen Shot 2022-04-05 at 12 47 53 pm

**[common v2.0.5]**
- Updates types to handle new chain parameters

**[injected-wallets v2.0.3]**
- Fixes chainId formatting issue on Binance wallet
- Refactors chainId interceptions for Binance and Coinbase wallet

### Checklist


packages/core/src/views/account-center/Index.svelte
```diff
@@ -0,0 +1,46 @@
+<script lang="ts">
+  import { onDestroy } from 'svelte'
+  import { updateAccountCenter } from '../../store/actions'
+  import type { AccountCenter } from '../../types'
+  import Maximized from './Maximized.svelte'
+  import Minimized from './Minimized.svelte'
+
+  export let settings: AccountCenter
+
+  const accountCenterPositions = {
```
Adamj1232, line 10: Would a centered option be useful?

aaronbarnardsound: Depends what you mean by customizable positioning? If you mean exact spacing from the top or right etc, I don't think it is worth having specific values there. We also need to know the position (ie topRight) so that we can do animations correctly as well for future UI elements that will be added either on top or beneath depending on the position.

Adamj1232 (approved):

packages/core/src/views/account-center/Index.svelte
Adamj1232: Totally makes sense

packages/core/README.md
```diff
@@ -235,15 +260,27 @@ Onboard currently keeps track of the following state:
...
 ```typescript
 type AppState = {
-  chains: Chain[]
   wallets: WalletState[]
+  chains: Chain[]
+  accountCenter: AccountCenter
   walletModules: WalletModule[]
 }
 
+type Chain {
```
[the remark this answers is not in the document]
aaronbarnardsound: Yeah good catch on this, I have updated the Chain on line 44 to have all of the params.

PR #919, opened 2022-04-05
fix: update brave wallet identity check

josheleonard:
### Description
Updates the identity check for Brave Wallet.

Screen Shot 2022-04-05 at 4 04 28 PM

### Checklist
- [X] The version field in `package.json` is incremented following [semantic versioning](https://semver.org/)
- [X] The box that allows repo maintainers to update this PR is checked
- [X] I tested locally to make sure this feature/fix works


aaronbarnardsound:
Hey @josheleonard, thank you for this PR.

Do you happen to know if this change is backwards compatible with older versions of the Brave browser. My understanding is that the `isBraveWallet` flag was not previously attached to the Brave provider, so was wondering if this would break users on older versions?

josheleonard:
@aaronbarnardsound, with this change, versions of Brave without the isBraveWallet variable will fallback to showing connect with MetaMask.

packages/demo/package.json
```diff
@@ -25,7 +25,7 @@
     "@web3-onboard/core": "^2.0.4",
     "@web3-onboard/fortmatic": "^2.0.0",
     "@web3-onboard/gnosis": "^2.0.0",
-    "@web3-onboard/injected-wallets": "^2.0.0",
+    "@web3-onboard/injected-wallets": "../injected",
```
aaronbarnardsound, line 28:
```suggestion
    "@web3-onboard/injected-wallets": "^2.0.3",
```

packages/injected/src/icons/brave.ts
```diff
@@ -1,20 +1,32 @@
...
+<svg width="218px" height="256px" viewBox="0 0 218 256" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
aaronbarnardsound, line 2:
```suggestion
<svg height="100%" viewBox="0 0 218 256" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
For all of our icons the largest dimension needs to be set to 100%, so in this case the height is larger, so we set that to 100% and then the width will scale automatically. This allows us to scale the icons for all of the UI.

aaronbarnardsound (changes requested):
Just a couple of small changes and then I think we will be good to go.

aaronbarnardsound (approved):
Looks good

aaronbarnardsound:
@josheleonard In what cases is the `isBrave` flag added to the provider? I am on the latest version and the provider does not have that flag.

josheleonard:
@aaronbarnardsound, currently the `isBraveWallet` provider flag is only present in Beta builds. This flag will be present in release builds soon.

Issue #932, opened 2022-04-11
[Bug]: magic wallet UX bugs


jadbox:
### Current Behavior

"@web3-onboard/magic": "^2.0.0",
"@web3-onboard/core": "^2.2.0",

Bug #1: The email sign in modal pops up BEHIND Onboard.js wallet prompt.
<img width="596" alt="Screenshot 2022-04-11 072046" src="https://user-images.githubusercontent.com/595843/162760923-19efabd3-f69d-4ddf-9686-d89201625f18.png">

Bug #2: On page reload, there should be a way to reconnect Magic without the user re-inputting their email. Ideally we should be able to store in localstorage and then provide to the Magic init constructor the email address to use on page reload.
<img width="526" alt="Screenshot 2022-04-11 072334" src="https://user-images.githubusercontent.com/595843/162761421-8c626b55-85fa-43d2-927d-0826bbc1b320.png">

### Expected Behavior

Bug #1) Magic email login modal should be above the Onboard UX
Bug #2) there should be a way to provide the Magic adaptor an email address to use to auto-reconnect on page reload

### Onboard Version

2.2.0

### Node Version

16

Adamj1232:
Hi @jadbox I am looking into this now. I will also expose the z-index of the login modal as an external variable...I have a PR [here](https://github.com/blocknative/web3-onboard/pull/934) where you can track the progress

Adamj1232:
Closing issue as this has been addressed in the latest Magic 2.0.1 release. Thank you for pointing this out! - https://github.com/blocknative/web3-onboard/releases/tag/web3-onboard%2Fmagic-v2.0.1

PR #935, opened 2022-04-11
[magic-v2.0.1] : Fix - Add userEmail init option to handle logged in session and add configurable z-index

Adamj1232:
### Description
This changes add a userEmail init option for the Magic package which will allow users to continue their session without having to log in again as well as an increase login modal z-index and expose as external variable to handle [928](https://github.com/blocknative/web3-onboard/issues/932)

### Checklist


packages/magic/src/validation.ts
```diff
@@ -0,0 +1,20 @@
+import Joi from 'joi'
+import type { MagicInitOptions } from './types'
+
+const magicOptions = Joi.object({
+  apiKey: Joi.string().required(),
+  userEmail: Joi.string()
```
aaronbarnardsound, line 6:
```suggestion
  userEmail: Joi.string().allow(null)
```
We could allow null here to make is easy for initialization since localStorage returns null if the value is not there. So the magic module could be initialized directly like this:

```javascript
const magic = magicModule({ apiKey: 'dsdfash3r43hf', userEmail: localStorage.getItem('magicUserEmail') })
```

packages/magic/README.md
```diff
@@ -28,7 +29,7 @@ type magicOptions = {
 import Onboard from '@web3-onboard/core'
 import magicModule from '@web3-onboard/magic'
 
-const magic = magicModule({ apiKey: 'API_KEY' })
+const magic = magicModule({ apiKey: 'API_KEY', userEmail: 'test@email.com' })
```
aaronbarnardsound, line 32:
```suggestion
const magic = magicModule({ apiKey: 'API_KEY', userEmail: localStorage.getItem('magicUserEmail') })
```
See below suggestion about updating validation so that the module can be initialized nice and easy like this.

aaronbarnardsound:
Since we have added the `Joi` package, maybe you could switch out the validation [here](https://github.com/blocknative/web3-onboard/pull/935/files#diff-1ff9b05ce3be323f88713c664e2b9a55068cf987a47f2aeeacb5b4d668814eb1L22-L27) to use `Joi.string().email()` validation?

packages/magic/README.md
```diff
@@ -43,13 +44,17 @@ console.log(connectedWallets)
...
+When a Magic wallet is connected the Magic instance is exposed. 
 This can be used to get information such as user MetaData, update a user's email address or handle the user's token.
+The user's email can be set in local storage and passed through the `MagicInitOptions` to avoid a user having to login again if they are revisting the DApp within the set user session time. 
+Magic has a default time of 7 days and this can be configured through your Magic API Key settings.
 ```typescript
 const [magicWallet] = await onboard.connectWallet()
 
 try {
   const { email, publicAddress } = await magicWallet.instance.user.getMetadata();
+  localStorage.set('magicUserEmail': email)
```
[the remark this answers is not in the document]
Adamj1232: lol duh, thank you!

Adamj1232:
> Since we have added the `Joi` package, maybe you could switch out the validation [here](https://github.com/blocknative/web3-onboard/pull/935/files#diff-1ff9b05ce3be323f88713c664e2b9a55068cf987a47f2aeeacb5b4d668814eb1L22-L27) to use `Joi.string().email()` validation?

Sweet! I have that implemented and was able to remove that ugly regex!

aaronbarnardsound (approved):
Just a suggested change as we probably don't want to allow

PR #939, opened 2022-04-12
[common-v2.0.6, core-v2.2.3, magic-v2.0.2] : Enhancement - Modal positioning variables and background color variable for main scroll container

Adamj1232:
### Description
Add modal positioning variables to main modal, account select and magic login modal

### Checklist


Adamj1232:
> For customizing the modal positioning, I wonder if this approach might be more flexible:
> 
>     * Set the parent full screen container to `relative` position
> 
>     * Set the modal position to `absolute``
> 
>     * Set the parent container to be flex centered vertically and horizontally (so the default is in the middle)
> 
>     * Add the following variables and their defaults:
>       `--modal-top: unset;`
>       `--modal-right: unset;`
>       `--modal-bottom: unset;`
>       `--modal-left: unset`;
> 
>     * Then the position can be set to wherever desired. For instance, the modal could be set 16px from the top but still centered horizontally: `--onboard-modal-top: 16px;`
> 
>     * No need for margins
> 
> 
> Let me know what you think!

Sounds great and cleaner! I have implemented that along with changes to the variables in the separate styled components. Let me know if that layout makes more sense

packages/core/src/views/shared/Modal.svelte
```diff
@@ -65,11 +72,16 @@
...
-  <div on:click={close} class="background flex justify-center items-center">
-    <div on:click|stopPropagation class="flex relative max-height">
-      <div class="modal-overflow relative flex justify-center">
-        <div class="modal relative">
-          <slot />
+  <div
+    on:click={close}
+    class="background flex items-center justify-center relative"
+  >
+    <div class="flex modal-position">
```
aaronbarnardsound, line 79:
```suggestion
    <div class="flex modal-position absolute">
```

aaronbarnardsound: This needs to be positioned absolutely so that the positioning variables work.

Adamj1232: Good catch!

aaronbarnardsound (approved):
Looks good!

PR #947, opened 2022-04-13
[core-v2.2.2] : Fix - Error arising after disconnecting wallet

Adamj1232:
### Description
Adds a check causing this error to be thrown when diconnecting a wallet
Screen Shot 2022-04-13 at 12 55 35 PM

### Checklist

### DIFF
--- a/packages/core/src/views/account-center/Minimized.svelte
+++ b/packages/core/src/views/account-center/Minimized.svelte
@@ -41,9 +41,7 @@
       : null
 
   $: primaryChain =
-    primaryWallet && primaryWallet.chains && primaryWallet.chains.length
-      ? primaryWallet.chains[0]
-      : undefined
+    primaryWallet && primaryWallet.chains[0]
 
   $: validAppChain = chains.find(({ id, namespace }) =>
     primaryChain
--- a/packages/core/src/views/account-center/Minimized.svelte
+++ b/packages/core/src/views/account-center/Minimized.svelte
@@ -40,7 +40,10 @@
       ? firstAccount.balance[firstAddressAsset]
       : null
 
-  $: primaryChain = primaryWallet.chains[0]
+  $: primaryChain =
+    primaryWallet && primaryWallet.chains && primaryWallet.chains.length
+      ? primaryWallet.chains[0]
+      : undefined
 
   $: validAppChain = chains.find(({ id, namespace }) =>
     primaryChain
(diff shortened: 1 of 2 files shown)


packages/core/src/views/account-center/Minimized.svelte
```diff
@@ -40,15 +40,18 @@
...
+      : undefined
 
   $: validAppChain = chains.find(({ id, namespace }) =>
     primaryChain
       ? id === primaryChain.id && namespace === primaryChain.namespace
       : false
   )
 
-  $: defaultChainStyles = getDefaultChainStyles(primaryChain.id)
+  $: defaultChainStyles = getDefaultChainStyles((primaryChain && primaryChain.id) ? primaryChain.id : undefined)
```
aaronbarnardsound, line 54:
```suggestion
  $: defaultChainStyles = getDefaultChainStyles(primaryChain && primaryChain.id)
```
Could this be shortened to the above since `primaryChain` would be `undefined` if it does not have the `id` property?

packages/core/src/views/account-center/Minimized.svelte
```diff
@@ -40,15 +40,18 @@
       ? firstAccount.balance[firstAddressAsset]
       : null
 
-  $: primaryChain = primaryWallet.chains[0]
+  $: primaryChain =
+    primaryWallet && primaryWallet.chains && primaryWallet.chains.length
+      ? primaryWallet.chains[0]
+      : undefined
```
aaronbarnardsound, line 46:
```suggestion
  $: primaryChain =
    primaryWallet && primaryWallet.chains[0]
```
Since `primaryWallet.chains` is always an array I think we can skip the `.chains` and `.length` checks

aaronbarnardsound (approved):
Looks good. Added a couple of suggestions that would need to be tested, but I think we can make it a littler more terse.

Also we should go ahead and increment the version to get this in to a release and then increment the React package web3-onboard version as well.

packages/core/src/views/account-center/Minimized.svelte
codingwithdidem: I'd be better this way

`primaryChain = primaryWallet?.chains?.[0]`

Adamj1232: @codingwithdidem thank you for the review and I agree the optional chaining is more clean but for wide range bundler support we are unable to include JS operators later than ES7

PR #953, opened 2022-04-15
[fix] HD Wallet signing for L2s

taylorjdawson:
### Checklist


packages/common/package.json
```diff
@@ -41,6 +41,7 @@
     "typescript": "^4.5.5"
   },
   "dependencies": {
+    "@ethereumjs/common": "2.6.2",
```
Adamj1232, line 44: I like moving the getCommon functionality out of the HD wallets, I am a little concerned with adding a ~2.5mb package to common

taylorjdawson: I think it's okay since it won't have any effect on bundle size and is dynamically imported so won't effect users unless they use the hdwallets

Adamj1232 (commented):
May want to bump the common version in the react package's and its version also so it stay up to date

aaronbarnardsound (approved):
Nice, looks good!

Adamj1232 (approved):
May want to bump the common version in the react package's package.json and its version also so it stay up to date otherwise looks good!

taylorjdawson:
> May want to bump the common version in the react package's and its version also so it stay up to date otherwise looks good!

@Adamj1232 what do you mean?

PR #961, opened 2022-04-21
Fixes type issue with keepkey

taylorjdawson:
### Checklist


Adamj1232 (approved):
I don't have a keepkey to test with but that looks spot on!

PR #962, opened 2022-04-21
Feature/populate txn

taylorjdawson:
### Checklist


Adamj1232 (approved):
Works well for trezor!

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/react",
-  "version": "2.1.7-alpha.2",
+  "version": "2.1.8-alpha.3",
```
Adamj1232, line 3:
```suggestion
  "version": "2.1.7-alpha.3",
```
Because there is already a `-alpha.x` tag on the version you wouldn't need to update the version number unless the previous version bump wasn't adequate in capturing the changes made - There was already a patch version update, if the changes you were adding required a minor version bump then the version would be updated.

taylorjdawson: got it! 👍🏾

PR #965, opened 2022-04-25
[fix]: Fixes signing issue with ledger

taylorjdawson:
### Description
There was an issue with signing messages in ledger where the recovered address didn't match the signing address. Just needed to remove the hex prefix before signing.

### Checklist

### DIFF
--- a/packages/ledger/src/index.ts
+++ b/packages/ledger/src/index.ts
@@ -126,8 +126,12 @@ function ledger({
           '@ethersproject/providers'
         )
 
-        const { accountSelect, createEIP1193Provider, ProviderRpcError, getCommon } =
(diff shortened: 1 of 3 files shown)


packages/ledger/src/index.ts
```diff
@@ -203,17 +207,17 @@ function ledger({
           const account =
             accounts.find(account => account.address === address) || accounts[0]
 
+          console.log('signMessage', { address, message, account })
```
Adamj1232, line 210: Remove this?

Adamj1232 (approved):
Nice! Works well, demo needs better UX eventually since there isnt any user feedback after `Sign Message` is clicked, took me a second to check the Ledger

PR #987, opened 2022-05-08
[fix] Duplicate coinbase wallets

taylorjdawson:
### Checklist

### DIFF
--- a/packages/coinbase/src/index.ts
+++ b/packages/coinbase/src/index.ts
@@ -1,10 +1,12 @@
 import { WalletInit } from '@web3-onboard/common'
 
-function coinbaseWallet(options?: { darkMode?: boolean }): WalletInit {
-  const { darkMode = false } = options || {}
-
-  return ({ device }) => {
-    return device.type === 'mobile'
+function coinbaseWallet({
+  darkMode = false
+}: {
+  darkMode?: boolean
+}): WalletInit {
+  return ({ device }) =>
+    device.type === 'mobile'
       ? {
           label: 'Coinbase',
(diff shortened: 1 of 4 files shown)


packages/coinbase/src/index.ts
```diff
@@ -3,51 +3,53 @@ import { WalletInit } from '@web3-onboard/common'
...
-        // patch the chainChanged event
-        const on = coinbaseWalletProvider.on.bind(coinbaseWalletProvider)
-        coinbaseWalletProvider.on = (event, listener) => {
-          on(event, val => {
-            if (event === 'chainChanged') {
-              listener(`0x${(val as number).toString(16)}`)
-              return
+  return ({ device }) => {
+    return device.type === 'mobile'
+      ? {
```
Adamj1232, line 8: Kind of weird to read as a ternary because the return is big but not a blocker

Adamj1232 (approved):

packages/coinbase/src/index.ts
taylorjdawson: yeah good point

taylorjdawson: I will make it more explicit by having the null at the top

Adamj1232: another approach would be to add a method guard at the top 
```suggestion
      if (device.type === 'mobile') return null
      return {
```
p.s. first time using suggestions button not sure how well it will work lol

taylorjdawson: The guard clause won't work like that as if it is mobile return the wallet otherwise return null -- but I removed some extra fluff so maybe it reads nicer?

PR #992, opened 2022-05-09
[core - v2.2.9] : Update - Device driven background scroll

Adamj1232:
### Description
Device driven background scroll

### Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.2.8",
+  "version": "2.2.9",
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
--- a/packages/core/src/views/shared/Modal.svelte
+++ b/packages/core/src/views/shared/Modal.svelte
@@ -15,7 +15,10 @@
 
 <script lang="ts">
   import { fade } from 'svelte/transition'
-  import { onDestroy, onMount } from 'svelte'
+  import { onDestroy, onMount } from 'svelte'  
+  import { getDevice } from '../../utils'
+  
+  const device = getDevice()
 
   const body = document.body
   onMount(() => {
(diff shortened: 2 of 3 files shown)


aaronbarnardsound:
@Adamj1232  we also need to remove the scroll listener when the modal component is destroyed to prevent memory leaks.

Adamj1232:
> 

Yep totally agree, was refactoring that last night. I would be interested in your's and @taylorjdawson opinions on targeting the HTML tag for controlling background positioning (mobile: `position`, desktop: `overflow`)
The reason for the different approaches is:
`overflow = 'hidden'` on mobile allows for a small amount of scrolling causing a gap at the bottom of the view port where the backdrop shadow doesn't cover
`position = 'fixed'` on desktop causes the page to jump to the top of the view port when the modal is opened
The approach of targeting the HTML for setting these styles is that the HTML tag is typically not a target for a lot of base styles that would cause side-effects as you saw with targeting the body element @taylorjdawson 
Open to ideas here

taylorjdawson:
@Adamj1232 what's the goal of this PR exactly?

taylorjdawson:
@Adamj1232 If I understand correctly you're saying to attach the modal as a child of the `<html />` tag rather than the body?

Adamj1232:
> 

The goal is to control the background positioning while the Onboard modal is open - I updated the description to reflect more thoroughly

gesquinca:
Another option to consider:

disable scroll: 
`document.ontouchmove = function(e){ e.preventDefault(); }`

enable scroll: 
`document.ontouchmove = function(e){ return true; }`

I believe we could use `scroll` for desktop browsers and `ontouchmove` on mobile browsers.

Adamj1232:
> 

I was really hoping to find a fix like you have that works across browsers but wasn't able to find anything reliable.
My thinking for this was if we can scope the css changes<Mobile-`position:fixed` & Desktop-`overflow:hidden`> to the html tag we will avoid interfering with most styling approaches

taylorjdawson (approved):

PR #1005, opened 2022-05-20
Add blockchain icons: heco moonbeam okx klaytn kcc hsc

chendatony31:
### Description

Add icons for heco moonbeam okx klaytn kcc hsc chains

### Checklist


chendatony31:
@Adamj1232  @aaronbarnardsound  Hi, could you pls have a look at this PR?

aaronbarnardsound:
Hey @chendatony31 I don't think that we will want to merge this PR. Adding an icon for every chain is not sustainable as it adds bloat to the package for every user regardless of whether they use that chain or not. We have a few select chain icons included in the package to make it easy for devs to support the most used chains out of the box.

If an app would like to include the above chains, they can configure them when initializing Onboard via the [`chains` parameter](https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard-develop/packages/common/src/types.ts#L424) by defining the `icon`.

An example with Avalanche would be:

```javascript
const onboard = Web3Onboard({
  //... other init params
  chains: [
    {
      id: '0xA86A',
      token: 'AVAX',
      label: 'Avalanche',
      rpcUrl: 'https://api.avax.network/ext/bc/C/rpc',
      icon: avalancheIconSvgString
    }
  ]
})
```

We are looking at creating a section in our docs where we can document examples of chain configurations with icons, so it makes it super easy for devs to copy and paste the chains they would like to support and once we have that setup, you can PR the chain icon in to the docs.

PR #915, opened 2022-04-02
refactor: upgrade to React 18 + add `useAppState`

iamacook:
### Description
This PR simplifies state subscription by using React 18 and the new `useSyncExternalStore` hook, as well as a few small tweaks.

1. Instead of manually subscribing to the state in each hook, a new `useAppState` hook has been added. It's possible to subscribe to the entire state or part of it via its argument, just as one would with the `select()` method.
2. The options argument of the `connect()` method has been made optional to match that of `connectWallet()`.
3. Unnecessary casting has been removed by destructuring methods from the `web3Onboard` instance directly.

I haven't altered the version number as I'm not sure how you guys would like to approach that seeing as it includes a substantial dependency upgrade.

### Checklist
- [ ] The version field in `package.json` is incremented following [semantic versioning](https://semver.org/)
- [x] The box that allows repo maintainers to update this PR is checked
- [ ] I tested locally to make sure this feature/fix works
- [ ] This PR passes the Circle CI checks

PR #1009, opened 2022-05-24
Update the react peerdep to not peg to React 17

ChristopherWirtOfficial:
This package is currently pegged by peer dependency to React 17, which is likely to be the least adopted React version in a while, given that it was a glorified minor version before React 18. This peer dep should be more than fine for what this package does, and just increases compatibility without causing problems for users that are now able to use the package where they wouldn't be allowed to before.

Currently this package _should_ be compatible with React 16, 17, and 18. But the Peer Deps listed basically disallow you from having any react packages that aren't in V17, which is almost certainly not the desired behavior, and a pretty annoying consequence of using the package currently. There's another PR (https://github.com/blocknative/web3-onboard/pull/915) that upgrades to React 18, but that isn't necessarily as compatible as this change and should be a lot easier to get through.

### Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -25,6 +25,6 @@
     "@web3-onboard/common": "^2.1.0-alpha.1"
   },
   "peerDependencies": {
-    "react": "^17.0.2"
+    "react": ">=16.8"
   }
 }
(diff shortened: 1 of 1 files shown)


ChristopherWirtOfficial:
Not needed now that #915 is merged

Issue #948, opened 2022-04-13
[Bug]: `addNewChain` should specify the blockchain explorer url


wrap-that-potassium:
### Current Behavior

`addNewChain` is missing the `blockExplorerUrls` setting, so the user has a created chain with **no** blockchain explorer.

### Expected Behavior

`addNewChain` should specify `blockExplorerUrls` setting, coming from enriched `chains` config when initializing `Onboard`.

### Steps To Reproduce

1. Start with a fresh MetaMask installation
2. Switch to a non included chain, like Polygon
3. Polygon network is added in MetaMask with **no** blockchain explorer url

### Onboard Version

2.2.1

### Node Version

16

aaronbarnardsound:
Hey @wrap-that-potassium that is a good suggestion and we can get that added.

mahmud-bn:
@wrap-that-potassium 
Closing this issue since [a solution has been merged ](https://github.com/blocknative/web3-onboard/pull/1021).

PR #1021, opened 2022-05-26
Feature: optional public RPC and block explorer Chain params

### LINKED ISSUE #1020: [Bug]: addNewChain always sets the app's provider RPC URL
### Current Behavior

In `onboard.setChain`, when the user doesn't have that chain in their wallet, onboard suggests to add it.

To achieve that, [it passes the app's provider configuration](https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard/packages/core/src/chain.ts#L71) to the wallet.

In a typical web3 app, that uses a custom Infura endpoint with an individual project id, it's not desirable that the user has this custom RPC URL in their wallet.

### Expected Behavior

There should be a way to specify an RPC URL to give to the wallet when adding a new chain. That would typically be a public RPC.

### Steps To Reproduce

* Use MetaMask
* Remove Polygon if you have it in your MetaMask already
* Initialize onboard with a custom Polygon RPC
* Call `onboard.setChain({ chainId: 137 })`
* Observe that it offers the user to install Polygon with that RPC

### Onboard Version

2

### Node Version

16

_No response_

### Relevant log output

_No response_

### Anything else?

I would suggest to add a new field in the Chain configuration, `publicRpc`, and use that when calling `wallet_addEthereumChain`. Or maybe just a new param for the `setChain` method.

katspaugh:
### Description
Resolves #1020 and #948.

As outlined in #1020, it's sometimes necessary to pass a custom (typically public) RPC when adding a new chain to the user's wallet.

### Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.2.13",
+  "version": "2.2.13-alpha.1",
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.2.12-alpha.1",
+  "version": "2.2.12-alpha.2",
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
@@ -41,7 +41,7 @@
     "typescript": "^4.5.5"
   },
   "dependencies": {
-    "@web3-onboard/common": "^2.1.1-alpha.1",
+    "@web3-onboard/common": "^2.1.1-alpha.2",
     "bowser": "^2.11.0",
     "ethers": "5.5.3",
     "eventemitter3": "^4.0.7",
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/common",
-  "version": "2.1.2",
+  "version": "2.1.2-alpha.1",
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/common",
-  "version": "2.1.1-alpha.1",
+  "version": "2.1.1-alpha.2",
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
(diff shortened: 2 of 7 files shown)


packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.2.9-alpha.4",
+  "version": "2.2.9-alpha.5",
```
katspaugh, line 3: I'm not sure which version would be correct here.
`2.2.9-alpha.5` or `2.2.10-alpha.4`?

packages/demo/src/App.svelte
```diff
@@ -269,7 +269,7 @@
     <button on:click={() => onboard.setChain({ chainId: '0x4' })}
       >Set Chain to Rinkeby</button
     >
-    <button on:click={() => onboard.setChain({ chainId: '0x89' })}
+    <button on:click={() => onboard.setChain({ chainId: '0x89', chainRpcUrl: 'https://polygon-rpc.com/' })}
```
katspaugh, line 272: I couldn't find any tests in this repo, so I updated the demo. It won't do anything unless the version of the core package is published and updated correspondingly in the demo package.json.

Please let me know if there's any test that I could extend instead.

katspaugh:
@aaronbarnardsound @Adamj1232 @taylorjdawson gm! Any feedback for this?

aaronbarnardsound:
Thanks for this PR @katspaugh. Another approach that I think might be more clear naming wise, is to add an optional `publicRpcUrl` parameter to the `Chain` object and use that in the add chain method. That way the `Chain` info can be set in one place and then there's no need to think about it when setting the chain. I also think it will help distinguish between the two `rpcUrl` parameters. What do you think?

aaronbarnardsound:
Just adding a note @taylorjdawson while we are here we should also add an option `blockExplorerUrl` parameter as requested in #948

katspaugh:
Makes sense!

I'll update the PR to extend the Chain object then 👍

taylorjdawson:
Hey @katspaugh! I see some commits here, is this ready for re-review?

katspaugh:
I haven't fully tested this version yet but a review would be appreciated, yes!

taylorjdawson:
@katspaugh This looks good! Have you had a chance to test?

katspaugh:
Still wrapping my head around testing monorepos, I'll get back to you on the weekend!

taylorjdawson:
@katspaugh I can also test if you want, was there anything else you wanted to add to the PR?

katspaugh:
Oh, that'd be awesome. Let me just fix those git conflicts in the package.json's.

mahmud-bn (approved):

mahmud-bn:
Hi @katspaugh 
Thanks for creating the PR. This has been merged

katspaugh:
Amazing, thank you guys!

PR #1047, opened 2022-06-08
fix(keystone): use let for AirGappedKeyring const

conwayconstar:
### Description
Use let assignment for AirGappedKeyring as its currently using const and then reassigning variable

### Checklist

### DIFF
--- a/packages/keystone/package.json
+++ b/packages/keystone/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/keystone",
-  "version": "2.1.0",
+  "version": "2.1.1",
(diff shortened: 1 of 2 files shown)


packages/keystone/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/keystone",
-  "version": "2.1.0",
+  "version": "2.1.1",
```
taylorjdawson, line 3:
```suggestion
  "version": "2.1.1-alpha.1",
```

taylorjdawson (changes requested):

taylorjdawson (approved):

PR #1059, opened 2022-06-13
add cndl to docs

samisbakedham:


taylorjdawson:
hey @samisbakedham thanks for the PR! We have examples for most popular chains and think that the ones we currently have in docs suffice in showing examples of how to use web3-onboard, so we'll keep it as is for now :)

samisbakedham:
> hey @samisbakedham thanks for the PR! We have examples for most popular chains and think that the ones we currently have in docs suffice in showing examples of how to use web3-onboard, so we'll keep it as is for now :) 

Hi, what qualifies as most popular chains?

PR #1083, opened 2022-06-20
[core-v2.3.0-alpha.2, react-v2.2.0-alpha.2, vue-v2.1.0-alpha.2] : Fix - Validation to handle onclick and links properly

Adamj1232:
### Description
Update Notifications validation to handle onclick and links properly

### Checklist

### DIFF
--- a/packages/core/src/validation.ts
+++ b/packages/core/src/validation.ts
@@ -189,7 +189,9 @@ const customNotificationUpdate = Joi.object({
   eventCode: Joi.string(),
   message: Joi.string().required(),
   id: Joi.string().required(),
-  autoDismiss: Joi.number()
+  autoDismiss: Joi.number(),
+  onclick: Joi.function(),
+  link: Joi.string()
 })
 
 const customNotification = Joi.object({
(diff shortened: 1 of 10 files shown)


packages/core/src/validation.ts
```diff
@@ -189,7 +189,9 @@ const customNotificationUpdate = Joi.object({
   eventCode: Joi.string(),
   message: Joi.string().required(),
   id: Joi.string().required(),
-  autoDismiss: Joi.number()
+  autoDismiss: Joi.number(),
+  onclick: Joi.function(),
```
taylorjdawson, line 193: @Adamj1232 maybe should have caught this sooner but can we make it camel case?? `onClick`

taylorjdawson (approved):
approve with nit

mahmud-bn (approved):

packages/core/src/validation.ts
Adamj1232: @taylorjdawson the only reason i didn't was that Notify v1 didn't use camel case and didn't want to break backwards compatibility. We can update but what are your thoughts there?

PR #1095, opened 2022-06-24
[docs]: Simplify Quickstart

aaronbarnardsound:
### Description
Modifies docs for Core, React and Vue packages so that the quick-start sections have the minimal code necessary to get up and running and also adds examples of how to create and Ethers provider with the connected wallet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,11 @@
 
 Install the core Onboard library, the injected wallets module and optionally ethers js to support browser extension and mobile wallets:
 
-`npm i @web3-onboard/core @web3-onboard/injected-wallets ethers`
+**NPM**
+`npm i @web3-onboard/vue @web3-onboard/injected-wallets ethers`
 
-- [@web3-onboard/core Official NPM Documentation](https://www.npmjs.com/package/@web3-onboard/core)
+**Yarn**
+`yarn add @web3-onboard/vue @web3-onboard/injected-wallets ethers`
 
 Then initialize in your app:
 
@@ -53,7 +55,7 @@ const ethersProvider = new ethers.providers.Web3Provider(
 
 const signer = ethersProvider.getSigner()
 
-// send a transaction with an ethers provider
+// send a transaction with the ethers provider
 const txn = await signer.sendTransaction({
   to: '0x',
   value: 100000000000000
--- a/README.md
+++ b/README.md
@@ -47,22 +47,24 @@ const wallets = await onboard.connectWallet()
 
 console.log(wallets)
 
-// create an ethers provider with the last connected wallet provider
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -71,7 +88,7 @@ For full documentation, check out the README.md for each package:
 - [Torus](packages/torus/README.md)
 - [WalletConnect](packages/walletconnect/README.md)
 - [WalletLink](packages/walletlink/README.md)
```
Adamj1232, line 90: Do we want to swap WalletLink out for Coinbase?

packages/react/README.md
```diff
@@ -2,188 +2,63 @@
...
-      transaction: {
-        txRequest: 'Su transacción está esperando que confirme'
-      }
-    }
-  }
+  ]
 })
 
 function App() {
   const [{ wallet, connecting }, connect, disconnect] = useConnectWallet()
```
Adamj1232, line 42: When I was writing docs for the new hooks I wasn't sure the best way to note hooks in the example code. I eventually went with importing all and only showing in action of a couple, with the thinking if they copy/paste a dev has them imported and ready but not a ton of extra code to deal with. I can also understand this approach but it requires more reading of the docs to get the other hooks which can be a challenge for some 🤷‍♂️

Adamj1232:
Looks good! Couple notes - and there is a conflict caused by the new react hooks docs 😬

Adamj1232 (approved):
Nice simplified docs!

taylorjdawson (approved):

mahmud-bn (approved):

Ia6368060:
Nice

packages/react/README.md
aaronbarnardsound: Yeah I agree, it is a bit of a trade off. We can start by simplifying the quick-start for now. I think in the longer term we probably want to rewrite the docs so that it includes a step by step walkthrough of how Onboard helps you solve common web3 app problems (connecting a wallet, tracking the connected addresses, switching networks), combined with a straight API that is derived from our types. That way we cover it in depth, or a dev can just take a quick look at the types for what they need.

README.md
aaronbarnardsound: Yeah good call.

README.md
```diff
@@ -12,17 +12,20 @@
 
 ## Quickstart
 
-Install the core Onboard library and the injected wallets module to support browser extension and mobile wallets:
+Install the core Onboard library, the injected wallets module and optionally ethers js to support browser extension and mobile wallets:
```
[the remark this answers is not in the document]
aaronbarnardsound: Yeah I guess I use them interchangeably, but will update.

Issue #1076, opened 2022-06-18
[Feature]: Pre-selection of network on WalletConnect session request


tfalencar:
### Is your request related to a problem?

With WalletConnect the Dapp tells the wallet which network to connect to on establishing the connection. 

Currently, when onboard library is configured, even if chainId of 1 is not even specified, a walletconnect session request is sent with chainId of '1' (Ethereum mainnet) instead of whatever is specified in the "chains" array. The problem with this approach is that then yet another request needs to be sent for the user to authorise "switching networks", which is quite bad UX. Modern / good mobile wallets allows simultaneous connections of different dapps to multiple networks at the same time, by keeping track of which dapp is connected to which chainId. The widespread "metamask standard" is bad even for browsers: imagine having 5 dapps in 5 tabs, and having to switch networks every-time a tab in the browser is switched - makes no sense at all, so it is good that walletconnect doesn't follow the metamask approach and we should allow this.

### Feature Description

As a user from the onboard library, I'd like to specify the networks my dapp supports like this:

const onboard = Onboard({
  wallets: [walletConnect],
  chains: [
    {
      id: '0xA',
      token: 'ETH',
      label: 'Optimism',
      rpcUrl: '....'
    }
    ]}
});

And have the walletconnect session request dialog on the mobile wallet side (iOS or Android) to already make the request to connect to the proper network. One solution for when there is support for multiple networks, is for Onboard to send the walletconnect session request with the first item in the array, as the "default" network for the request.

### Alternative Solutions

Here's another dapp describing a similar issue: https://github.com/gnosis/cowswap/issues/1973 

### Anything else?

Screenshot of the desired result on the mobile wallet side (as example, rainbow wallet was used):
<img width="421" alt="Screenshot 2022-06-17 at 15 27 48" src="https://user-images.githubusercontent.com/4054009/174437714-658d5d07-8250-467d-bf59-c07e6c529f95.png">

aaronbarnardsound:
Just did some digging in to this and I believe that we can set a `chainId` when connecting the session by changing [this line here](https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard-develop/packages/walletconnect/src/index.ts#L132) to : `this.connector.createSession({chainId: parseInt(chains[0].id, 16)})` and that would connect to the first chain in the `chains` array passed in on init. We will need to check what WalletConnect does on that call if the chain is not supported by the wallet and handle that accordingly.

tfalencar:
[...]

PR #1099, opened 2022-06-24
[walletconnect-v2.0.4-alpha.2] : Feature - Connect First ChainId on WalletConnect session request

mahmud-bn:
### Description

PR in response to Issue #1076
### Checklist

### DIFF
--- a/packages/demo/src/App.svelte
+++ b/packages/demo/src/App.svelte
@@ -220,25 +220,9 @@
 
     const signature = await signer.signTransaction({
       to: '',
-      value: 100000000000000
+      value: 1000000000000000
     })
-
     console.log(signature)
-  }
-
-  let toAddress
-  const sendTransaction = async (provider) => {
-    const ethersProvider = new ethers.providers.Web3Provider(provider, 'any')
-
-    const signer = ethersProvider.getSigner()
-
-    const txn = await signer.sendTransaction({
-      to: toAddress,
-      value: 100000000000000
-    })
-
-    const receipt = await txn.wait()
-    console.log(receipt)
   }
 
   const signMessage = async (provider, address) => {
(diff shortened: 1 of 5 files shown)


packages/walletconnect/src/index.ts
```diff
@@ -129,7 +130,7 @@ function walletConnect(options?: WalletConnectOptions): WalletInit {
                   // Check if connection is already established
                   if (!this.connector.connected) {
                     // create new session
-                    this.connector.createSession().then(() => {
+                    this.connector.createSession(connectFirstChainId ? {chainId: parseInt(chains[0].id, 16)} : undefined).then(() => {
```
taylorjdawson, line 133: @mahmud-bn Just double checking here. Do we need to add some check to make sure that `chains` and `chains[0]` are defined? 
For instance:
```suggestion
                    this.connector.createSession(connectFirstChainId && chains && chains[0] ? {chainId: parseInt(chains[0].id, 16)} : undefined).then(() => {
```

taylorjdawson (changes requested):
Can we bump walletconnect in the demo package.json as well to `2.0.4-alpha.2`

packages/demo/src/App.svelte
```diff
@@ -114,38 +114,7 @@
       gnosis,
       dcent
     ],
-    chains: [
```
taylorjdawson, line 117: @mahmud-bn Is there a reason we want to get rid of these chains?

packages/walletconnect/src/index.ts
taylorjdawson: @mahmud-bn Awesome! Just wanted to double check

packages/demo/src/App.svelte
mahmud-bn: Oh! Fixed that. I was just re-checking the chain error.

taylorjdawson (approved):

aaronbarnardsound (approved):

PR #1100, opened 2022-06-27
Update BUG.yml

taylorjdawson:
* Adds Adam and Mahmud to assignees
* Adds drop down to select effected package
* Makes node version optional 
* Adds note to include build config and repro steps
* Adds dropdown to select between build and runtime issue
* Adds sanity check at the end of the issue to make sure user submits issue with build config or repro steps


Adamj1232 (approved):
Nice!!

PR #1102, opened 2022-06-28
2.0.3-alpha.2: [fix] - MEW Set Chain

### LINKED ISSUE #1044: [Bug]: MEW module cannot change network on desktop and cannot connect on mobile
### Current Behavior

I have application with web3-onboard setup that is working perfect for injected wallets, coinbase and walletconnect. As they are working, I assume that my configuration of `Onboard` object is fine.

Then I wanted to add MEW module via just adding next wallet in array like here https://docs.blocknative.com/onboard/mew-my-ether-wallet. What issues I've got:

Issue 1: After chosing MEW in connect modal, I can scan the QR code with my mobile MEW wallet app and application connect perfectly, There is no option to pick network in mobile application after connection. When I am trying to change network with `setChain` method (which is working for other wallets that i have in wallets array) nothing happened and I see only `400 error on POST to rpcUrl` in console.

Issue 2: When I am trying to enter hosted application through MEW wallet mobile app and their dapp browser, and I pick MEW method in modal, mobile app is asking for picking one of wallets, but then onboard modal is still in `connecting` state. Also what is interesting, in this embeded mode, modal detects Trust wallet available. After chosing it, bahavior is the same (connecting state never ends).

### Expected Behavior

Issue 1: Since MEW mobile app doesn't allow to change network in UI after connection, it would be perfect to change network via `setChain`.

Issue2: On mobile connection also should be working - it can be connected to Trust as it detects it in this mode.

### Steps To Reproduce

Issue1:
1. Add mew module to working setup like here https://docs.blocknative.com/onboard/react-hooks
2. Connect MEW mobile wallet throught qr scanning
3. Try to trigger `setChain` method

Issue2:
1. Enter MEW mobile wallet
2. Enter `dapp browser` there
3. Pick an url of app that have added mew module 
4. Pick MEW and if you have more than 1 wallet in MEW app (chose proper one).

### Onboard Version

@web3-onboard/mew - 2.0.0

### Node Version

16.10.0

Chrome

### Relevant log output

_No response_

### Anything else?

I was trying to change rpcUrl in config to wss, but other wallets were throwing errors, even if MEW was not raising `POST 400`, but still nothing happened on `setChain`.

aaronbarnardsound:
### Description
The MEW provider does not support the `wallet_addEthereumChain` and `wallet_switchEthereumChain`, so this PR sets those fields to null when creating the provider, which will throw an unsupported error when called.

Fixes #1044 

### Checklist


taylorjdawson (changes requested):
@aaronbarnardsound I think we should re-init mew on chain switch (if possible) rather than just not supporting the operation

aaronbarnardsound:
@taylorjdawson I believe that MEW only supports mainnet - https://myetherwallet.github.io/MEWconnect-Protocol-Documentation/mewconnect_provider.html#methods
Check the `chainId` parameter.

PR #1115, opened 2022-06-30
[core: 2.3.2-alpha.4]: [fix] -  Recommended Injected Wallets

aaronbarnardsound:
### Description
When only the injected wallets module is used and the user has no injected wallets installed, Onboard was no longer displaying the links included in the `reccomendedInjectedWallets` parameter of the `appMetadata` object. This was due to checks not being updated for `autoSelect` now being an object that is always set.

This PR updates all checks for the `autoSelect` parameter which fixes this issue.

### Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.3.2-alpha.3",
+  "version": "2.3.2-alpha.4",
   "repository": "blocknative/web3-onboard",
   "scripts": {
     "build": "rollup -c",
--- a/packages/core/src/views/connect/Index.svelte
+++ b/packages/core/src/views/connect/Index.svelte
@@ -396,7 +396,7 @@
                   {connectingErrorMessage}
                 />
               </div>
-            {:else if !walletToAutoSelect}
+            {:else}
               <InstallWallet />
             {/if}
           {/if}
--- a/packages/core/src/views/connect/Index.svelte
+++ b/packages/core/src/views/connect/Index.svelte
@@ -52,7 +52,7 @@
   let scrollContainer: HTMLElement
 
   let walletToAutoSelect =
-    autoSelect &&
+    autoSelect.label &&
     walletModules.find(
       ({ label }) => label.toLowerCase() === autoSelect.label.toLowerCase()
     )
@@ -209,7 +209,7 @@
       // user rejected account access
       if (code === ProviderRpcErrorCode.ACCOUNT_ACCESS_REJECTED) {
         connectionRejected = true
-        if (autoSelect) {
+        if (walletToAutoSelect) {
           walletToAutoSelect = null
 
           if (autoSelect.disableModals) {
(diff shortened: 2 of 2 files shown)


taylorjdawson (approved):
noice! 🚀

mihaic195:
Hey @aaronbarnardsound  I'm getting `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'label')` on line 55. Could this be a bug?

aaronbarnardsound:
Hey @mihaic195, could you share how you are calling the `connectWallet` function?

aaronbarnardsound:
@mihaic195 I get the error you described if an empty object is passed in to the connect wallet function: `onboard.connectWallet({})`.

Is that what you are passing in?

Our validation should not allow that value, so I am going to update it to error indicating that is not a valid value.

aaronbarnardsound:
Fix is included in #1121

ArshiaDhiman:
Hey @aaronbarnardsound I'm getting this error: TypeError: Cannot read properties of undefined (reading 'label'), is this a bug? I am putting an empty object in the connect({}) when a user clicks the connect wallet button and using the web3-onboard/react. Also use the autoSelect options when trying to connect using a useEffect and staying connected when the page is refreshed. 

 connect({
                autoSelect: {
                    label: previouslyConnectedWallets[0],
                    disableModals: true,
                },
            }).then(setCurrentChain);

Have been stuck on this a while just wondering if maybe I am using it wrong? Thanks

aaronbarnardsound:
Hey @ArshiaDhiman yes passing in an empty object is what causes the error and the fix above fixes our validation to indicate that an empty object is an invalid parameter to the `connectWallet` function. If you are not auto selecting a wallet, just call the `connectWallet` function without any arguments (`connectWallet()`).

ArshiaDhiman:
Okay perfect thanks @aaronbarnardsound. I just thought I would specify that I am using @web3-onboard/react. I was unable to call the connect() function without a empty object but using the code below I fixed the error.


         `connect({
                                autoSelect: {
                                label:  "any",
                                disableModals: false,
                            },
                        }).then(setChain)`

aaronbarnardsound:
Ah ok, we need to get the types fixed in the React package then! I'll create an issue so that we can get that fixed up.

aaronbarnardsound:
@ArshiaDhiman Actually it looks like our types are correct and the connect options are optional - https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard-develop/packages/react/src/index.ts#L74

ArshiaDhiman:
@aaronbarnardsound okay thanks. I still get a error when I try use connect without a empty object or autoSelect. I have attached an example here, the fix above solved my issue so there is no problem. Thanks! 😄   Screenshot (394)

aaronbarnardsound:
@ArshiaDhiman which version of the react package are you on? Could you try updating to the latest?

ArshiaDhiman:
The issue was fixed thanks for your help

PR #1119, opened 2022-07-01
[core: 2.4.0-alpha.1] - [enhancement] - Set Primary Account

aaronbarnardsound:
### Description
This PR exposes a new action to the public API that allows for setting the primary (first in the list) wallet and optionally the account if multiple accounts are connected for a particular wallet. This function was used internally for the Account Center, but has now been moved to the `actions.ts` file so that it can be exposed.

Other minor inclusions in this PR:
- `OnboardAPI` has been modified to being the `typeof` of the exposed API. This will save us having to update the types every time we add a new method to it.
- Some labels and custom error messages have been added to some of the Joi validations to ensure more useful error messages.
- Updates docs

### Checklist


packages/core/src/index.ts
```diff
@@ -35,14 +40,16 @@ const API: OnboardAPI = {
...
       customNotification,
       updateBalances,
-      updateAccountCenter
+      updateAccountCenter,
+      setPrimaryWallet
     }
   }
 }
 
+export type OnboardAPI = typeof API
```
Adamj1232, line 49: @aaronbarnardsound I like this update! Should we use this type in React/Vue packages instead of the OnboardAPI type in the types folder?

aaronbarnardsound: That type is exported from this file and then imported by the React and Vue packages, so it already is!

Adamj1232: Ah I see, perfect! 🤘🏼

Adamj1232 (approved):
Nice!!

sgurmani:
@aaronbarnardsound great work indeed. Can you please let me know when it is realising ?

aaronbarnardsound:
@sgurmani No worries, will do. I am aiming for a release in the next couple of days.

PR #1135, opened 2022-07-08
[common-v2.1.5-alpha.1, core-v2.4.0-alpha.7, react-v2.2.3-alpha.5, vue-v2.1.3-alpha.5, demo-v2.0.6]: Enhancement -  chain number support

mahmud-bn:
### Checklist


packages/core/src/index.ts
```diff
@@ -86,7 +87,7 @@ function init(options: InitOptions): OnboardAPI {
   } = options
 
   initI18N(i18n)
-  addChains(chains)
+  addChains(hexedChain(chains))
```
taylorjdawson, line 90:
```suggestion
  addChains(chains.map(chainIdToHex))
```

packages/core/src/index.ts
```diff
@@ -26,6 +26,7 @@ import {
 } from './store/actions'
 
 import updateBalances from './update-balances'
+import { hexedChain } from './utils'
```
taylorjdawson, line 29:
```suggestion
import { chainIdToHex } from './utils'
```

packages/core/src/utils.ts
```diff
@@ -92,6 +92,16 @@ export async function copyWalletAddress(text: string): Promise<void> {
...
+export function hexedChain(chains : Chain[]): Chain[] {
+  const hexChain = chains.map(({ id, ...rest }) => { 
+    if(typeof id === 'number'){
+       id = '0x'+Math.abs(id).toString(16)
+      }
+      return { id, ...rest };
+  });
+  return hexChain
+}
+
```
taylorjdawson, line 104:
```suggestion
export const chainIdToHex = (chain: Chain): Chain => typeof id === 'number' ? { ...chain, id: `0x${chain.id.toString(16)}`} : chain
```
- When possible opt for ternary over multiple returns as I find it more clear/concise
- Template strings are usually preferred over concatenation
- Negative values should just fail so I don't think we need the absolute value
- Also I think that `chainIdToHex` is more reusable and follows similar pattern/naming convention as the object below 👍🏾

packages/core/src/validation.ts
```diff
@@ -21,7 +21,9 @@ import type {
   Notify
 } from './types'
 
-const chainId = Joi.string().pattern(/^0x[0-9a-fA-F]+$/)
+// const chainId = Joi.string().pattern(/^0x[0-9a-fA-F]+$/)
+const chainId = Joi.alternatives().
+                try(Joi.string().pattern(/^0x[0-9a-fA-F]+$/), Joi.number())
```
taylorjdawson, line 26:
```suggestion
const chainId = Joi.alternatives().
                try(Joi.string().pattern(/^0x[0-9a-fA-F]+$/), Joi.number().positive())
```

taylorjdawson: Here we can enforce positivity with joi ☝🏾 🙌🏾

taylorjdawson: @aaronbarnardsound @Adamj1232 Is there a reason for the duplicated validation code in `packages/common/src/validation.ts` (seen above)  and here? and is there a way to consolidate to make it DRYer?

taylorjdawson (changes requested):

packages/core/src/utils.ts
mahmud-bn: Hmmn. Thanks. Adapted with a tiny addition.

taylorjdawson (approved):
Looks great! Versioning is okay as long as #1137 this gets merged first.

PR #1137, opened 2022-07-10
[common-v2.1.7-alpha.1, core-v2.6.0-alpha.1, react-v2.2.5-alpha.1, vue-2.1.5-alpha.1, demo-v2.0.6]: Enhancement - Convert Wei to Eth

mahmud-bn:
### Checklist


packages/core/src/utils.ts
```diff
@@ -83,6 +84,13 @@ export function shortenEns(ens: string): string {
   return ens.length > 11 ? `${ens.slice(0, 4)}...${ens.slice(-6)}` : ens
 }
 
+export function weiToEth(wei: any): string {
+  const bigWei = new BigNumber(wei._hex)
+  const bigWeiDiv = bigWei.dividedBy(1e18)
+  const formattedWei = bigWei.dividedBy(1e18).toFixed() === '0' ? bigWeiDiv.toFixed(1) : bigWeiDiv.toFixed()
+  return formattedWei
+}
```
aaronbarnardsound, line 92:
```suggestion
export function weiToEth(wei: any): string {
  return new BigNumber(wei).div(1e18).toString()
}
```

aaronbarnardsound: Not sure what the `_hex` refers to here. This function should take a string value in `wei` as input and return the string in `ether` without the full decimal places. Then where ever it is being rendered, it can be decided how many decimals render.

aaronbarnardsound (changes requested):

packages/core/src/utils.ts
mahmud-bn: logging wei will give you `BigNumber {_hex: '0x05443223fb7515db', _isBigNumber: true}`. Returning `new BigNumber(wei).div(1e18).toString()` without accessing _hex will give you `NaN`.

aaronbarnardsound: That would mean that the function is being called with a Bignumber instance rather than a string. My suggestion would be if possible, to ensure the function is always called with a string value for consistency.

packages/common/src/hdwallets.ts
```diff
@@ -71,11 +71,11 @@ export const bigNumberFieldsToStrings = (
     (transaction, txnProperty) => ({
       ...transaction,
       ...((transaction[txnProperty as keyof TransactionRequest] as BigNumber)
-        .toHexString
+        .toString(16)
```
aaronbarnardsound, line 74: It looks like this first expression is not meant to call the `toString` method and it is actually just checking if it is a big number or not. You might be able to use an `instanceof` check instead to see if the value is a Bignumber or not.

packages/common/src/utils.ts
```diff
@@ -0,0 +1,5 @@
+import type BigNumber from 'bignumber.js'
+
+export function weiToEth(wei: BigNumber): string {
+    return wei.div(1e18).toString()
```
aaronbarnardsound, line 4:
```suggestion
    return wei.div(1e18).toString(10)
```
We should put a radix of `10` here to ensure decimal formatting and not exponential notation.

packages/core/src/provider.ts
```diff
@@ -339,14 +339,15 @@ export async function getBalance(
...
+  const { wallets } = state.get()
 
   try {
-    const balanceWei = await provider.getBalance(address)
-
-    return balanceWei
-      ? { [chain.token || 'eth']: utils.formatEther(balanceWei) }
+    const wallet = wallets.find(wallet => wallet.label)
+    const provider = wallet.provider
+    const balanceHex = await provider.request({ method: 'eth_getBalance', params:[address,'latest'] });
```
aaronbarnardsound, line 348: We are going to want a decimal balance here rather than hex.

packages/core/src/provider.ts
```diff
@@ -339,14 +339,15 @@ export async function getBalance(
...
+    return balanceHex
+      ? { [chain.token || 'eth']: shortenHexToString(balanceHex) }
```
aaronbarnardsound, line 350: I don't think we want to shorten the balance string here and if we were to do that we should probably round the number. The balance should just return the full balance as a decimal string. Then wherever the balance gets rendered, it can be decided how many decimal places should be rendered.

packages/core/src/utils.ts
```diff
@@ -83,6 +83,11 @@ export function shortenEns(ens: string): string {
   return ens.length > 11 ? `${ens.slice(0, 4)}...${ens.slice(-6)}` : ens
 }
 
+export function shortenHexToString(hex: string) : string {
+  const hexToString = (parseInt(hex, 16) / 1e18).toString()
+  return hexToString.length > 5 ?  hexToString.slice(0,6) : hexToString
+}
+
```
aaronbarnardsound, line 90: I think that this can be removed as it is not needed. Also not sure what it is exactly doing? Seems to be doing more than just shortening the string with that divide in there.

aaronbarnardsound (changes requested):

packages/common/src/utils.ts
mahmud-bn: Thanks. I was thinking that was the default. 
`If the radix is not specified, the preferred radix is assumed to be 10.`

packages/core/src/provider.ts
mahmud-bn: Hmmn. Makes sense. So I reverted it back to the original ethers BigNumber "BalanceWei' version but using 'bignumber.js' instead.

packages/core/src/provider.ts
```diff
@@ -345,9 +347,10 @@ export async function getBalance(
   try {
     const wallet = wallets.find(wallet => wallet.label)
     const provider = wallet.provider
-    const balanceHex = await provider.request({ method: 'eth_getBalance', params:[address,'latest'] });
-    return balanceHex
-      ? { [chain.token || 'eth']: shortenHexToString(balanceHex) }
+    const balanceHex = await provider.request({ method: 'eth_getBalance', params:[address,'latest'] })
+    const balanceWei = new BigNumber(parseInt(balanceHex, 16))
+    return balanceWei
+      ? { [chain.token || 'eth']: weiToEth(balanceWei).slice(0,6) }
```
aaronbarnardsound, line 353: I am still not sure why we need to `slice` here? Was that in the original code? I think we should return the full decimal here and then let it be truncated or rounded where ever it is rendered. We pass this balance to all subscriptions, so dapps will most likely want the full value and then decide what to do with it.

packages/common/src/hdwallets.ts
```diff
@@ -71,11 +71,11 @@ export const bigNumberFieldsToStrings = (
     (transaction, txnProperty) => ({
       ...transaction,
       ...((transaction[txnProperty as keyof TransactionRequest] as BigNumber)
-        .toHexString
+        instanceof BigNumber
```
aaronbarnardsound, line 74: I did some testing and since these values are coming from the populate transaction method in Ethers, it will be the Ethers instance of BigNumber which use BN.js, so the `instanceof` check will not work. So I think we can actually revert this to the original code.

packages/core/src/provider.ts
```diff
@@ -339,14 +341,16 @@ export async function getBalance(
...
   if (!chain) return null
-
-  const provider = getProvider(chain)
+  
+  const { wallets } = state.get()
 
   try {
-    const balanceWei = await provider.getBalance(address)
-
+    const wallet = wallets.find(wallet => wallet.label)
```
taylorjdawson, line 348: @mahmud-bn does this just return the first wallet that has the label property? 
also what happens if `wallet` is `undefined`? 
maybe to make it more explicit do something like:
```
const wallet = wallets.find(wallet => !!wallet.label)
```
or even?
```
const wallet = wallets.find(wallet => !!wallet)
```

taylorjdawson: Or even could do it on the provider which would also check that the provider is defined:
```
const wallet = wallets.find(wallet => !!wallet.provider)
```

packages/common/src/utils.ts
```diff
@@ -0,0 +1,5 @@
+import type BigNumber from 'bignumber.js'
+
+export function weiToEth(wei: BigNumber): string {
+    return wei.div(1e18).toString(10)
+}
```
taylorjdawson, line 5:
```suggestion
}

```

packages/common/src/index.ts
```diff
@@ -126,4 +128,4 @@ export type {
   CustomNetwork,
   TransactionObject,
   RPCResponse
-}
+}
```
taylorjdawson, line 131:
```suggestion
}

```

taylorjdawson (changes requested):
Looks good! Just a few suggestions

packages/core/src/provider.ts
mahmud-bn: @taylorjdawson 
- Yes, it does. Calling the getBalance() function gets the updated balance of all connected wallets.
- Since getBalance() is called after the wallet connection has taken place, I assume that was enough protection. But I've added the provider check suggestion.

packages/common/src/utils.ts
mahmud-bn: @taylorjdawson 
- Is there a reason why I should place the extra linespace in line 6?

taylorjdawson (approved):

aaronbarnardsound (approved):

PR #1138, opened 2022-07-11
[core-v2.4.0-alpha.8, react-v2.2.3-alpha.5, vue-v2.1.3-alpha.6]  : Feature - Preflight notification handling

Adamj1232:
### Description
Adds Pre-flight Notification functionality to deliver standard notifications along with preflight information by passing a `PreflightNotificationOptions` object to the `preflightNotification` action. This will return a a promise that resolves to the transaction hash or null `Promise<string> | null`.

Preflight event types include 
 - `txRequest` : Alert user there is a transaction request awaiting confirmation by their wallet
 - `txAwaitingApproval` : A previous transaction is awaiting confirmation
 - `txConfirmReminder` : Reminder to confirm a transaction to continue - configurable with the `txApproveReminderTimeout` property; defaults to 15 seconds
 - `nsfFail` : The user has insufficient funds for transaction (requires `gasPrice`, `estimateGas`, `balance`, `txDetails.value`)
 - `txError` : General transaction error (requires `sendTransaction`)
 - `txSendFail` : The user rejected the transaction (requires `sendTransaction`)
 - `txUnderpriced` : The gas price for the transaction is too low (requires `sendTransaction`)

```typescript
interface PreflightNotificationOptions {
  sendTransaction?: () => Promise<string>
  estimateGas?: () => Promise<string>
  gasPrice?: () => Promise<string>
  balance?: string | number
  txDetails?: {
    value: string | number
    to?: string
    from?: string
  }
  txApproveReminderTimeout?: number // defaults to 15 seconds if not specified <confirmed default with Gustavo>
}
```

### Checklist


packages/core/src/index.ts
```diff
@@ -39,6 +40,7 @@ const API = {
       setLocale,
       updateNotify,
       customNotification,
+      preflightNotification,
```
aaronbarnardsound, line 43: I wonder if the function should be called `preflightNotifications` instead to indicate that it renders a series of notifications rather than just one?

packages/core/src/preflight-notifications.ts
```diff
@@ -0,0 +1,242 @@
...
+  notificationsArr = notifications
+})
+
+export async function preflightNotification(
+  options: PreflightNotificationOptions
+): Promise<string> | null {
+  const { apiKey } = configuration
+
+  if (!apiKey) {
+    console.error(
```
aaronbarnardsound, line 25: I don't think an apiKey is required for the preflight notifications as the Blocknative server is not accessed right?

packages/core/README.md
```diff
@@ -637,6 +637,63 @@ setTimeout(
...
+ - `txAwaitingApproval` : A previous transaction is awaiting confirmation
+ - `txConfirmReminder` : Reminder to confirm a transaction to continue - configurable with the `txApproveReminderTimeout` property; defaults to 15 seconds
+ - `nsfFail` : The user has insufficient funds for transaction (requires `gasPrice`, `estimateGas`, `balance`, `txDetails.value`)
+ - `txError` : General transaction error (requires `sendTransaction`)
+ - `txSendFail` : The user rejected the transaction (requires `sendTransaction`)
+ - `txUnderpriced` : The gas price for the transaction is too low (requires `sendTransaction`)
+
+```typescript
+interface PreflightNotificationOptions {
+  sendTransaction?: () => Promise<string>
```
aaronbarnardsound, line 654: It might not be necessary for the `sendTransaction` function to resolve to a hash, since Notify is not using the hash for tracking post flight transactions. We could make the return type just be `Promise<void>`, where the promise resolves when the user confirms the transaction and rejects when the user rejects the transaction. Not sure if that makes things simpler or not, but sometimes depending on how the transaction is being sent, devs might need to jump through extra hoops to get it to resolve to the transaction hash.

aaronbarnardsound (changes requested):
Just a couple of comments, but otherwise looks pretty good 👍

packages/core/src/index.ts
Adamj1232: Good point! Ill update

packages/core/src/preflight-notifications.ts
Adamj1232: yep good point as well for the preflights, updating

packages/core/README.md
aaronbarnardsound: Yeah that makes sense.

aaronbarnardsound (approved):
Looks good 👍

PR #1145, opened 2022-07-14
[core: 2.5.0-alpha.1] - [enhancement] - Delayed Wallet Init

aaronbarnardsound:
### Description
This PR changes the timing of when the wallet initialization code is run from during Onboard init, to when the `connectWallet` function is first called. This will allow for all async injected wallet modules to be injected in the page before we initialize the wallet modules. This important due to changes enforced to Google extensions in early 2023 which will make all injected wallets have to inject async. 
In the case of an autoselect wallet flow, we also add an additional small delay to ensure enough time for injection.

### Checklist


taylorjdawson (approved):
Nice looks good to me! 🚀

PR #1158, opened 2022-07-21
common-v2.1.7-alpha.2, core-v2.6.0-alpha.3, react-v2.2.5-alpha.2, vue-v2.1.5-alpha.2]: Fix: Chain id type fix

mahmud-bn:
### Description

Fixes chain type errors for other packages

### Checklist


packages/core/src/chain.ts
```diff
@@ -8,7 +8,7 @@ import { validateSetChainOptions } from './validation'
 import type { WalletState } from './types'
 
 async function setChain(options: {
-  chainId: string | number
+  chainId: string
```
aaronbarnardsound, line 11:
```suggestion
  chainId: string | number
```
So I think that we still want to accept a `number` type here for the `chainId` param.

Then under this line: `const { chainId, chainNamespace = 'evm', wallet: walletToSet } = options`
we want to normalize it as a hex string: 
```javascript
const chainIdHex = typeof chainId === 'number' ? `0x${chainId.toString(16)}` : chainId
```

Then we can refer to `chainIdHex` instead of `chainId` for all comparisons underneath.

You could change the utils functions so that we have a simple function that takes a chainId and ensures it is a hex string and then have another function that takes an array of chains and uses that function to convert all chainId's to hex.

packages/core/src/utils.ts
```diff
@@ -92,7 +93,13 @@ export async function copyWalletAddress(text: string): Promise<void> {
   }
 }
 
-export const chainIdToHex = (chain: Chain): Chain => typeof chain.id === 'number' ? { ...chain, id: `0x${chain.id.toString(16)}` } : chain
+export function chainIdToHex(chains : Chain[] | ChainWithDecimalId[] ): 
+Chain[] {
+  return chains.map(({ id, ...rest }) => { 
+    id = typeof id === 'number' ?  `0x${Math.abs(id).toString(16)}` : id
```
aaronbarnardsound, line 99: Why the need for `Math.abs`? I think if we want to ensure that it is a positive number, we should add that to the validation with the `positive` function.

aaronbarnardsound (changes requested):
Some requested changes and then I can do some tests locally.

packages/core/src/utils.ts
```diff
@@ -93,12 +93,14 @@ export async function copyWalletAddress(text: string): Promise<void> {
   }
 }
 
+export const decimalToHex = (decimal: number): string => `0x${decimal.toString(16)}`
```
aaronbarnardsound, line 96: Ok so to keep things DRY, we can create a function that will take a number or string and ensure that it is a hex string:

```typescript
export const toHexString = (val: number | string): string => typeof val === 'number' ? `0x${decimal.toString(16)}` : val
```

Then we can use that in the `chainIdToHex` func:

```typescript
export function chainIdToHex(chains : Chain[] | ChainWithDecimalId[] ): 
Chain[] {
  return chains.map(({ id, ...rest }) => { 
    const hexId = toHexString(id)
    return { id: hexId, ...rest }
  })  
} 
```

We can then replace line 24 in src/chain.ts with:

```javascript
const chainIdHex = toHexString(chainId)
```

packages/core/src/chain.ts
```diff
@@ -20,15 +21,16 @@ async function setChain(options: {
...
   // validate that chainId has been added to chains
   const chain = chains.find(
-    ({ namespace, id }) => namespace === chainNamespace && id === chainId
+    ({ namespace, id }) => namespace === chainNamespace && id === chainIdHex
   )
 
   if (!chain) {
     throw new Error(
-      `Chain with chainId: ${chainId} and chainNamespace: ${chainNamespace} has not been set and must be added when Onboard is initialized.`
+      `Chain with chainId: ${chainIdHex} and chainNamespace: ${chainNamespace} has not been set and must be added when Onboard is initialized.`
```
aaronbarnardsound, line 33:
```suggestion
      `Chain with chainId: ${chainId} and chainNamespace: ${chainNamespace} has not been set and must be added when Onboard is initialized.`
```
We can use the original chainId for this message to prevent confusion, as the dev passes in a decimal and the error message indicates that they passed in a hex.

aaronbarnardsound (changes requested):

aaronbarnardsound (approved):

Adamj1232 (approved):

PR #1165, opened 2022-07-25
Fix trezor nonce

bachstatter:
### Description
This PR resolves an issue were we pass a nonce missing 0x prefix. 

To get the transaction signed on line 404 in [trezor/src/index.ts](https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard-develop/packages/trezor/src/index.ts#L395) we call `Transaction.fromTxData`. The `Transaction` class is coming from `@ethereumjs/tx`.
When calling `fromTxData` that package will instantiate a class extending their `BaseTransaction`. The `BaseTransaction` constructor will try to convert the `nonce` to first a `buffer` and then `BigInt` [see here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/baseTransaction.ts#L101)  . To convert the string to a buffer they rely on `toBuffer` from `@ethereumjs/util`, this method will throw if passed a string not prefixed with `0x` [See here]( https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/bytes.ts#L149)

Using an infura provider I consistently got a nonce without `0x`. I also fixed some minor type errors.

### Checklist

### DIFF
--- a/packages/trezor/package.json
+++ b/packages/trezor/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/trezor",
-  "version": "2.1.7-alpha.1",
+  "version": "2.1.7-alpha.2",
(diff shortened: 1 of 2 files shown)


bachstatter:
Should I do something to trigger the Circle CI checks?

Adamj1232:
> Should I do something to trigger the Circle CI checks?

@bachstatter no worries on the CI checks for now. We are working on adding those back into our new architecture but they do not currently run on branches

PR #1176, opened 2022-07-27
(main) Release: 2.6.0

Adamj1232:
### ✨ Highlights

-  Fix: Autoselect with disabled modals resolve #1154 
-  Fix: Chain id type fix #1158 
-  Enhancement: Minify bundle size #1155 
-  Fix: add "main" to Keystone's package.json #1164 
-  Fix Trezor nonce handling #1165 
-  Fix - Add optionality to containerElement, format validation so accountCener #1166 
-  Feature : Gas Module #1168 
-  Fix - Updated Binance Wallet provider patch #1174 
-  Enhancement : Convert internal Wei to Eth #1137 
-  Feature : Add BitKeep injected Wallet #1177 

### 📦 Changes per package

- `@web3onboard/coinbase` **v2.0.10**
- `@web3onboard/common` **v2.1.7**
- `@web3onboard/core` **v2.6.0**
- `@web3onboard/dcent` **v2.0.7**
- `@web3onboard/formatic` **v2.0.9**
- `@web3onboard/gas` **v2.0.0**
- `@web3onboard/gnosis` **v2.0.8**
- `@web3onboard/injected` **v2.0.15**
- `@web3onboard/keepkey` **v2.1.7**
- `@web3onboard/keystone` **v2.1.8**
- `@web3onboard/ledger` **v2.1.7**
- `@web3onboard/magic` **v2.0.10**
- `@web3onboard/mew` **v2.0.7**
- `@web3onboard/portis` **v2.0.7**
- `@web3onboard/react` **v2.2.5**
- `@web3onboard/torus` **v2.0.8**
- `@web3onboard/trezor` **v2.1.7**
- `@web3onboard/vue` **v2.1.5**
- `@web3-onboard/walletconnect` **v2.0.8**
- `@web3onboard/web3auth` **v2.0.6**

### Checklist
- [ ] The version field in `package.json` is incremented following [semantic versioning](https://semver.org/)
- [ ] The box that allows repo maintainers to update this PR is checked
- [ ] I tested locally to make sure this feature/fix works
- [ ] I have run `yarn type-check` & `yarn build` to confirm there are not any associated errors
- [ ] This PR passes the Circle CI checks

PR #1180, opened 2022-07-28
[common-v2.1.7-alpha.5, core-v2.6.0-alpha.8, react-v2.2.5-alpha.7]fix: weifix

mahmud-bn:
### Description

This PR: 
- enforces the weiToETH function to accept only strings from core and common 
- returns ethers.BigNumber for GetInterfaceHelpers used by Ledger

### Checklist

### DIFF
--- a/packages/common/src/validation.ts
+++ b/packages/common/src/validation.ts
@@ -9,9 +9,9 @@ const basePaths = Joi.array().items(basePath)
 
 const chain = Joi.object({
   namespace: Joi.string(),
-  id: Joi.alternatives()
-        .try(Joi.string().pattern(/^0x[0-9a-fA-F]+$/), 
-        Joi.number().positive()).required,
+  id: Joi.string()
+    .pattern(/^0x[0-9a-fA-F]+$/)
+    .required(),
   rpcUrl: Joi.string().required(),
   label: Joi.string().required(),
   token: Joi.string().required(),
--- a/packages/core/src/provider.ts
+++ b/packages/core/src/provider.ts
@@ -350,7 +350,7 @@ export async function getBalance(
     const balanceHex = await provider.request({ method: 'eth_getBalance', params:[address,'latest'] })
     const balanceWei = new BigNumber(parseInt(balanceHex, 16))
     return balanceWei
-      ? { [chain.token || 'eth']: weiToEth(balanceWei) }
+      ? { [chain.token || 'eth']: weiToEth(balanceWei.toString()) }
       : null
   } catch (error) {
     console.error(error)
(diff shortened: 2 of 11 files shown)


packages/common/src/validation.ts
```diff
@@ -9,9 +9,9 @@ const basePaths = Joi.array().items(basePath)
 
 const chain = Joi.object({
   namespace: Joi.string(),
-  id: Joi.alternatives()
-        .try(Joi.string().pattern(/^0x[0-9a-fA-F]+$/), 
-        Joi.number().positive()).required,
+  id: Joi.string()
+    .pattern(/^0x[0-9a-fA-F]+$/)
+    .required(),
```
aaronbarnardsound, line 14: Any reason to change this validation?

packages/core/src/provider.ts
```diff
@@ -350,7 +350,7 @@ export async function getBalance(
     const balanceHex = await provider.request({ method: 'eth_getBalance', params:[address,'latest'] })
     const balanceWei = new BigNumber(parseInt(balanceHex, 16))
     return balanceWei
-      ? { [chain.token || 'eth']: weiToEth(balanceWei) }
+      ? { [chain.token || 'eth']: weiToEth(balanceWei.toString()) }
```
aaronbarnardsound, line 353: Since we are passing in a string, we could just pass in the `balanceHex` value directly to the `weiToEth` function.

aaronbarnardsound (changes requested):

packages/core/src/provider.ts
mahmud-bn: Yeah. Figured out introducing the BigNumber was an overkill. Thanks for pointing it out.

packages/common/src/validation.ts
mahmud-bn: This was related to the [validation error fix for hardware wallets chain added to the main release: ](https://github.com/blocknative/web3-onboard/pull/1176/commits/7b1eb11799663711d3594217b85e49fcbabdb8dd)

There is another Joi.alternatives- string or number [validation included in the previous PR core chainID which .](https://github.com/blocknative/web3-onboard/pull/1135/files#diff-e8cb2da76902122de27c23ccd59551757403f6fba1a2fb3c272c47096166c507R26)

I guess placing chains: Chain[] | ChainWithDecimalId[] will allow for such validation but I'm not sure that's needed since the GetInterfaceHelpers used by Ledger works with the internal hex string.

Adamj1232: This seem correct if we are using Hex chain IDs internally(converting number chain ids to hex after initialization) it doesn't seem we would need positive number validation within the common package

aaronbarnardsound:
Do we also need to remove the bignumber.js dep from core and the rollup externals since it is not being used anymore?

mahmud-bn:
> Do we also need to remove the bignumber.js dep from core and the rollup externals since it is not being used anymore?

It's still used by preflight-notifications.ts and notify.ts

Adamj1232 (approved):

aaronbarnardsound (approved):

PR #1182, opened 2022-07-31
Adds react context provider and nextjs example

taylorjdawson:
### Checklist


examples/with-nextjs/public/vercel.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="283" height="64" viewBox="0 0 283 64" fill="none"
```
Adamj1232, line 1: @taylorjdawson do we want to remove this?

examples/with-nextjs/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@web3-onboard/react": "file:.yalc/@web3-onboard/react",
```
mahmud-bn, line 12: I guess this is leftover from testing.

taylorjdawson: Yeah it is, good catch!

packages/react/package.json
```diff
@@ -50,20 +50,23 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "yarn clean && tsc",
+    "clean": "rimraf dist",
+    "prepublishOnly": "yarn build",
```
Adamj1232, line 55:
```suggestion
    "prepublish-only": "yarn build",
```
Kabob case to conform with other commands?

examples/with-nextjs/public/vercel.svg
Adamj1232: Not required but always nice to remove unnecessary files/code

packages/react/package.json
taylorjdawson: Oh that's the yarn lifecycle hook that gets called during yarn publish script and I think it need to be camel case?

PR #1191, opened 2022-08-02
[core-v2.7.0-alpha.1, react-v2.2.6-alpha.1, vue-v2.1.6-alpha.1] : Feature - Add configuration to hide/show sidebar

Adamj1232:
### Description
This PR adds the ability to add a configuration to hide the Connect Modal side bar.
The changes adds a `connect` property on the AppState and init options with an internal prop of `showSidebar : boolean` that defaults to `true`

Screen Shot 2022-08-02 at 5 16 34 PM

### Checklist
- [ ] I have run `yarn type-check` & `yarn build` to confirm there are not any associated errors


packages/core/src/validation.ts
```diff
@@ -184,6 +185,9 @@ const initOptions = Joi.object({
   gas: Joi.object({
     get: Joi.function().required(),
     stream: Joi.function().required()
+  }),
+  connect: Joi.object({
+    showSidebar: Joi.boolean()
```
aaronbarnardsound, line 190:
```suggestion
  connect: connectOptions
```

Could we bring the `connectOptions` up a few lines and then reuse it here?

packages/core/src/types.ts
```diff
@@ -274,6 +284,11 @@ export type UpdateAccountCenterAction = {
   payload: AccountCenter | Partial<AccountCenter>
 }
 
+export type UpdateConnectModalAction = {
+  type: 'update_connect_modal'
+  payload: ConnectModalOptions | Partial<ConnectModalOptions>
```
aaronbarnardsound, line 289:
```suggestion
  payload: Partial<ConnectModalOptions>
```

I know we have this pattern elsewhere, but I am pretty sure that we can simplify to just using the `Partial` as that just makes all parameters optional in the object, but would still allow for the full type to be passed there.

aaronbarnardsound (approved):
Just a couple of suggestions, but nothing crucial to the PR.

packages/core/src/validation.ts
Adamj1232: Oh yeah duh! Thanks man!

PR #1195, opened 2022-08-04
[core:2.7.0-alpha.2] [ledger:2.1.8-alpha.1] - [feature] - Replacement Transactions

aaronbarnardsound:
### Description
This PR adds a new feature to transaction notifications where if the notification is hovered, a new UI will dropdown to allow the user to speedup or cancel the transaction if the following conditions are met:

- The [gas](https://github.com/blocknative/web3-onboard/tree/v2-web3-onboard-develop/packages/gas) module has been passed in on initialization (required to estimate gas accurately for replacement transaction)
- The transaction was initiated on the networks that the gas module can get estimations (currently Ethereum and Polygon mainnet)
- The transaction is pending (`eventCode: 'txPool'`)
- The transaction was initiated by a hardware wallet

We are only supporting hardware wallets for this feature currently as we know they will support replacing transactions with the same nonce. Currently many wallets such as MetaMask ignore the nonce field of a transaction object.

Also included in this PR is a small bugfix for the Ledger module where connection errors were not propagating correctly, resulting in the wrong error message being displayed.

Since this feature cannot be tested with testnet funds, I have [deployed a demo](https://hw-replacement-tx-rinkeby.netlify.app/) here that can be tested on the Rinkeby network. To test:
- connect a hardware wallet
- Send a transaction by pasting in an address and clicking the "Send Transaction" button. This will send a transaction with low gas values.
- Once you see a pending notification for the transaction, hover on the notification and click "speedup" or "cancel" to replace the transaction.
- Confirm replacement transaction on your hardware wallet

### Checklist


packages/core/README.md
```diff
@@ -176,6 +176,12 @@ export type Notify = {
     event: EthereumTransactionData
   ) => TransactionHandlerReturn
   position: CommonPositions
+  replacement?: {
+    gasPriceProbability: {
+      speedup: number // default 80
```
Adamj1232, line 181: Might be helpful to add info on what the `speedup` and `cancel` props do in the configuration

packages/core/src/views/notify/Notification.svelte
```diff
@@ -2,32 +2,149 @@
...
+        removeTransaction(notification.id)
       }, notification.autoDismiss)
     }
   })
 
   onDestroy(() => {
     clearTimeout(timeoutId)
   })
+
+  function actionableEventCode(eventCode: Notification['eventCode']): boolean {
```
Adamj1232, line 58: @aaronbarnardsound Might be worth moving these functions into the notify.ts file or creating a new ts file for them like we did we preflight-notifications.ts...maybe even put all of them in a folder together to keep it clean...?

Adamj1232 (approved):
Looks good! A couple comments but nothing is a blocker

packages/core/src/views/notify/Notification.svelte
```diff
@@ -2,32 +2,149 @@
...
+    switch (wallet.label) {
+      case 'Ledger':
+      case 'Trezor':
+      case 'Keystone':
+      case 'KeepKey':
+        return true
+      default:
+        return false
+    }
+  }
```
taylorjdawson, line 89: @aaronbarnardsound maybe a little cleaner/simpler? 

```suggestion
const WALLETS_SUPPORT_REPLACEMENT = ['Ledger', 'Trezor', 'Keystone', 'KeepKey']
const walletSupportsReplacement = () => wallet && wallet.label in WALLETS_SUPPORT_REPLACEMENT
```

taylorjdawson (approved):
@aaronbarnardsound approved with nit!

packages/core/README.md
aaronbarnardsound: Yeah good call, added some more info.

packages/core/src/views/notify/Notification.svelte
aaronbarnardsound: Yeah agreed. Moved all of this logic out into a new `replacement.ts` file.

packages/core/src/views/notify/Notification.svelte
aaronbarnardsound: I have simplified based on this idea and have removed my cherished `switch` statements. 

By the way the `in` operator does not work when checking whether a `string` value is in an `array`. It is more of an object parameter lookup.

aaronbarnardsound:
Updates include:
- Moved the replacement logic out into it's own file
- Removed switch statements
- Added error handling with a hint notification if the replacement transaction fails
- Added more detail in to docs

packages/core/README.md
```diff
@@ -176,6 +185,14 @@ export type Notify = {
     event: EthereumTransactionData
   ) => TransactionHandlerReturn
   position: CommonPositions
+  replacement?: {
+    gasPriceProbability?: {
+      // define the gas price used for speedup based on the probability of getting in to the next block. Default 80
+      speedup?: number
+      // // define the gas price used for cancel based on the probability of getting in to the next block. Default 95
```
Adamj1232, line 192:
```suggestion
      // define the gas price used for cancel based on the probability of getting in to the next block. Default 95
```

Adamj1232 (approved):
Looks good! Tiny notes change but that's all!

PR #1205, opened 2022-08-10
Feature/sequence wallet

SamueleA:
### Description
This features is an integration of the [Sequence wallet](https://sequence.xyz/)
The wallet can be accessed either as a chrome extension or through the SDK. The former integration is done through the injected wallet package and the latter through the Sequence package.
If `window.ethereum.isSequence` is not true, the connection through the SDK will be made available since this will mean that the user does not have the chrome extension.

Notes: 
- signing message is not working on the demo app due to lack of support for EIP-1271 since Sequence uses a contract wallet.
The sequence SDK provider a method sequence.utils.isValidMessageSignature which allows the signature to be verified,
- switching to a network whose contract wallet is undeployed doesn't work sometimes/has a weird flow. This is a known issue which will be fixed on Sequence's side

### Checklist

### DIFF
--- a/packages/dcent/src/index.ts
+++ b/packages/dcent/src/index.ts
@@ -110,6 +110,7 @@ function dcent({
 
           const provider = new StaticJsonRpcProvider(currentChain.rpcUrl)
 
+          // @ts-ignore
           return generateAccounts(dcentKeyring, provider)
         }
 
--- a/packages/dcent/src/index.ts
+++ b/packages/dcent/src/index.ts
@@ -110,7 +110,6 @@ function dcent({
 
           const provider = new StaticJsonRpcProvider(currentChain.rpcUrl)
 
-          // @ts-ignore
           return generateAccounts(dcentKeyring, provider)
         }
 
--- a/packages/demo/webpack.config.js
+++ b/packages/demo/webpack.config.js
@@ -82,5 +82,9 @@ module.exports = {
     static: {
       directory: path.join(__dirname, 'public')
     }
+  },
+  performance: {
+    maxEntrypointSize: 3000000,
+    maxAssetSize: 3000000
   }
 }
--- a/packages/sequence/README.md
+++ b/packages/sequence/README.md
@@ -11,6 +11,7 @@
 ```typescript
 type SequenceOptions {
   appName?: string
+  network?: number | string
 }
 ```
 
--- a/packages/sequence/README.md
+++ b/packages/sequence/README.md
@@ -1,6 +1,6 @@
 # @web3-onboard/sequence
 
-## Wallet module for connecting Portis wallet to web3-onboard
+## Wallet module for connecting Sequence wallet to web3-onboard
 
 ### Install
 
--- a/packages/sequence/README.md
+++ b/packages/sequence/README.md
@@ -1,0 +1,37 @@
+# @web3-onboard/sequence
+
+## Wallet module for connecting Portis wallet to web3-onboard
+
+### Install
+
+`npm i @web3-onboard/sequence`
+
+## Options
+
+```typescript
+type SequenceOptions {
+  appName?: string
+}
+```
+
+## Usage
+
+```typescript
+import Onboard from '@web3-onboard/core'
+import sequenceModule from '@web3-onboard/sequence'
+
+const sequence = sequenceModule({
+  appName: 'My app'
+})
+
+const onboard = Onboard({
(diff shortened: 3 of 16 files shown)


aaronbarnardsound:
Hey @SamueleA thanks for the PR, we will try and get this reviewed soon.

aaronbarnardsound:
Hey @SamueleA I am taking a look at this at the moment.

> switching to a network whose contract wallet is undeployed doesn't work sometimes/has a weird flow. This is a known issue which will be fixed on Sequence's side

The wallet seems to default to Polygon, should I be able to switch to Ethereum mainnet, or does come under the known issue that you mention above? I see the wallet trigger a popup, but it sits there for a bit with a blank screen and then goes away without switching to Ethereum mainnet.

SamueleA:
> 

Yes, that is indeed the known issue I have mentioned. The Mainnet wallet is not deployed by default, just the polygon one. So the expected behavior when switching to Mainnet would have been a flow for that scenario. Instead, it popup just closes

SamueleA:
> Hey @SamueleA I am taking a look at this at the moment.
> 
> > switching to a network whose contract wallet is undeployed doesn't work sometimes/has a weird flow. This is a known issue which will be fixed on Sequence's side
> 
> The wallet seems to default to Polygon, should I be able to switch to Ethereum mainnet, or does come under the known issue that you mention above? I see the wallet trigger a popup, but it sits there for a bit with a blank screen and then goes away without switching to Ethereum mainnet.

Yes, that is indeed the known issue I have mentioned. The Mainnet wallet is not deployed by default, just the polygon one. So the expected behavior when switching to Mainnet would have been a flow for that scenario. Instead, it popup just closes

packages/sequence/README.md
```diff
@@ -0,0 +1,38 @@
+# @web3-onboard/sequence
+
+## Wallet module for connecting Portis wallet to web3-onboard
```
ihorbond, line 3: was this supposed to say Sequence ? :)

SamueleA: haha
Thanks for pointing it out! :)

packages/dcent/src/index.ts
```diff
@@ -110,6 +110,7 @@ function dcent({
 
           const provider = new StaticJsonRpcProvider(currentChain.rpcUrl)
 
+          // @ts-ignore
```
aaronbarnardsound, line 113: Were you getting type errors here? I am not seeing any type errors on this line that requires ts-ignore

packages/demo/webpack.config.js
```diff
@@ -82,5 +82,9 @@ module.exports = {
     static: {
       directory: path.join(__dirname, 'public')
     }
+  },
+  performance: {
+    maxEntrypointSize: 3000000,
+    maxAssetSize: 3000000
```
aaronbarnardsound, line 88: Was this needed for anything in particular?

packages/sequence/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  return () => {
+    // @ts-ignore
+    return window?.ethereum && window.ethereum?.isSequence ?
+      []
+    :
+     {
+      label: 'Sequence',
+      getIcon: async () => (await import('./icon.js')).default,
+      getInterface: async ({ chains }) => {
```
aaronbarnardsound, line 19:
```suggestion
      getInterface: async () => {
```
We can remove the unused reference.

aaronbarnardsound (changes requested):
Just a couple of small changes and questions.

packages/dcent/src/index.ts
SamueleA: I am indeed getting a type error there. Also, when building the package the dcent package produced an error that prevented the build process from succeeding.
But maybe it's just on my machine this is occurring since it's working with yours. I removed that line.

aaronbarnardsound (approved):

packages/sequence/src/index.ts
aaronbarnardsound: Thanks!

packages/demo/webpack.config.js
aaronbarnardsound: Cool makes sense 👍

packages/dcent/src/index.ts
aaronbarnardsound: Ah yeah that's weird. I'll take another look at that as a separate issue.

PR #1258, opened 2022-09-07
[injected-v2.2.2-alpha.1] : Fix - Add functionality to MM getInterface to check provider against all other provider flags

Adamj1232:
### Description
Add functionality to MM getInterface to check provider against all other provider flags.
This is needed in cases such as usage within Brave browser where Brave wallet has both a flag for `isMetaMask` and `isBraveWallet`
This added functionality can be added to other wallet interfaces in the case there is collision between provider flags.

### Checklist


packages/injected/src/wallets.ts
```diff
@@ -23,23 +23,43 @@ declare const window: CustomWindow
...
-  identity: string
+  identity: string,
+  checkOtherProviderFlags?: boolean
 ): () => Promise<{ provider: EIP1193Provider }> {
   return async () => ({
     provider: (window.ethereum.providers &&
     Array.isArray(window.ethereum.providers)
-      ? window.ethereum.providers.find(provider => !!provider[identity])
+      ? getInterfaceFromProvidersArray(identity, checkOtherProviderFlags)
       : window.ethereum) as EIP1193Provider
```
mario-iliev, line 33: I'm kind of guessing here since I don't know your code but it seems to me that the fix you did is only in case of coinbase (having provider.providers) but not for a single instance. In this case if you have MetaMask only in a Brave browser as you know Brave have a setting what should happen with the override. But in case I think I saw the behaviour where if you try to connect to Brave Wallet it is a success, then if you try to connect to MetaMask it is success again but in the end you receive two instances of the same wallet (Brave) in the "connectedWallets" array. After refresh is only one though.
Lon story short I really think you should check for other providers flags in every case.

In the codebase I work on I have something like:
`
const checkIfProviderIsCorrect = {
  'MetaMask': provider => Boolean(provider && provider.isMetaMask && !provider.isBraveWallet),
  'Brave Wallet': provider => Boolean(provider && provider.isBraveWallet),
  'Coinbase Wallet': provider => Boolean(provider && provider.isCoinbaseWallet),
};
`

This could be extended in the future and done in a more elegant way but you get the idea.
Cheers!

Adamj1232: @mario-iliev Thanks for the feedback.  
I don't foresee this being a problem as Brave Wallet and MetaMask inject into the same window space. 
Without Coinbase the user will only either see MetaMask or Brave (generally MetaMask as they inject into the `window.ethereum` space later thus overriding the Brave Wallet).
I have thoroughly tested and the fix is working as expected in situations where: 
- CB wallet + MM & Brave are enabled
- MM & Brave without CB wallet
- MM & Brave without CB wallet and `autoSelect` set to MetaMask (with different Brave wallet settings for the `window.ethereum` handling`).

If you would like you can give it a try.

taylorjdawson (approved):

taylorjdawson (approved):

aaronbarnardsound (approved):

Issue #1257, opened 2022-09-07
[Bug]: Wallet errors not propagated to onboard promises


mario-iliev:
### Current Behavior

Hey guys. I noticed something which makes handling errors impossible and lead to wrong/bad UX.
I only tested this with MetaMask and I can't confirm for other wallets but I suspect it's the same.
When using `onboard.connectWallet({ autoSelect: { label: 'MetaMask', disableModals: true }, })`

**Case 1:** User choose to connect with MetaMask, MM extension window is promped but user clicks outside of it so it gets hidden automatically. If the user tries to connect again, MetaMask fires this error `MetaMask - RPC Error: Already processing eth_requestAccounts. Please wait. {code: -32002, message: 'Already processing eth_requestAccounts. Please wait.'}` but the `onboard.connectWallet` Promise is not resolved so we can't show anything to the user and we actually don't know the current status of the "connectWallet" method.

**Case 2:** User choose to connect with MetaMask, MM extension is promped but user decides to cancel the connection. MM will fire this error `MetaMask - RPC Error: User rejected the request. {code: 4001, message: 'User rejected the request.'}` and the `onboard.connectWallet` will **resolve** with the currently connected wallets instead of rejecting with an error.

### Expected Behavior

Wallet errors should propagate to web3-onboard Promises which caused them.

### Steps To Reproduce

Steps to reproduce are described in the "Current Behavior" section.

### What package is effected by this issue?

@web3-onboard/core

### Is this a build or a runtime issue?

Runtime

### Package Version

2.8.1

### Node Version

16.14.2

### Sanity Check

- [X] If this is a build issue, I have included my build config. If this is a runtime issue, I have included reproduction steps and/or a [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example).

taylorjdawson:
Hey @mario-iliev! Thank you for bringing this to our attention! Does this issue only occur for you when using: `onboard.connectWallet({ autoSelect: { label: 'MetaMask', disableModals: true }, })` or does it also occur if simply calling: `onboard.connectWallet()`?

mario-iliev:
> Hey @mario-iliev! Thank you for bringing this to our attention! Does this issue only occur for you when using: `onboard.connectWallet({ autoSelect: { label: 'MetaMask', disableModals: true }, })` or does it also occur if simply calling: `onboard.connectWallet()`?
[...]

PR #1279, opened 2022-09-20
[core-v2.8.5-alpha.1, react-v2.3.5-alpha.1, vue-v2.2.5-alpha.1] : Fix - Handle add new chain canceled by user

Adamj1232:
### Description
Fixes: #1257 
Upon network switch, if the chain is not available in the wallet(MM) the user is prompted to add the chain, if the user cancels this action the promise now resolves to false.

### Checklist

### DIFF
--- a/packages/core/src/chain.ts
+++ b/packages/core/src/chain.ts
@@ -1,6 +1,6 @@
-import { firstValueFrom } from 'rxjs'
+import { BehaviorSubject, firstValueFrom } from 'rxjs'
 import { filter, map } from 'rxjs/operators'
-import { ProviderRpcErrorCode } from '@web3-onboard/common'
+import { Chain, ProviderRpcErrorCode } from '@web3-onboard/common'
 import { addNewChain, switchChain } from './provider.js'
 import { state } from './store/index.js'
 import { switchChainModal$ } from './streams.js'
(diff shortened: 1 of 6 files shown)


packages/core/src/chain.ts
```diff
@@ -66,21 +66,12 @@ async function setChain(options: {
...
-      try {
-        await addNewChain(wallet.provider, chain)
-        await switchChain(wallet.provider, chainIdHex)
-        return true
-      } catch (error) {
-        // display notification to user to switch chain
-        switchChainModal$.next({ chain })
-        return firstValueFrom(switchChainModalClosed$)
-      }
+      chainNotInWallet(wallet, chain, switchChainModal$, chainIdHex)
```
aaronbarnardsound, line 74: Do we need a `return` statement here?
```suggestion
      return chainNotInWallet(wallet, chain, switchChainModal$, chainIdHex)
```

packages/core/src/chain.ts
```diff
@@ -93,4 +84,28 @@ async function setChain(options: {
...
+  }>,
+  chainIdHex: string
+) => {
+  try {
+    await addNewChain(wallet.provider, chain)
+    await switchChain(wallet.provider, chainIdHex)
+    return true
+  } catch (error) {
+    const { code } = error as { code: number }
+    if (code === ProviderRpcErrorCode.ACCOUNT_ACCESS_REJECTED) {
```
aaronbarnardsound, line 101: Is it the same code for account access rejection and new chain rejection?

aaronbarnardsound (commented):
Just a couple of questions before approving.

packages/core/src/chain.ts
aaronbarnardsound: Yeah I think it is all good as is, just wanted to double check!

aaronbarnardsound (approved):
Looks good 👍

PR #1282, opened 2022-09-21
add meta tags to homepage 

gesquinca:
### Checklist
- [ ] I have run `yarn check` & `yarn build` to confirm there are not any associated errors


docs/src/routes/__layout-homepage.svelte
```diff
@@ -25,7 +25,7 @@
   }
 
   let title = 'Web3-Onboard Docs'
-  let description = 'Web3-Onboard Documentation'
+  let description = 'Open-source, framework-agnostic JavaScript library to onboard users to web3 apps. Help your users transact with ease by enabling wallet connection, real-time transaction states, and more.'
```
Adamj1232, line 28:
```suggestion
  const description = 'Open-source, framework-agnostic JavaScript library to onboard users to web3 apps. Help your users transact with ease by enabling wallet connection, real-time transaction states, and more.'
```
Doesn't look like this is re-assigned so it can be `const` instead of `let`

gesquinca: makes sense, thanks!

PR #1283, opened 2022-09-21
Precommit for not leaking secrets

aratz-lasa:
### Description
This PR adds [pre-commit](https://pre-commit.com/) for checking for secrets before doing a commit. This way we ensure developers do not leak their secrets. 

For avoiding false positive it is necessary to add a comment saying `pragma: allowlist secret` whenever a fake or example key is written. In this PR I added those comments, however I'm not sure if the comments I added are correct since I have not experience with Svelte templates and so.  

Moreover, I added hooks in pre-commit for checking trailing whitespaces and EOFs. So this PR some files were also cleaned.


Adamj1232:
Closing for now as we are moving to gh actions soon. Will def revisit this concept after transition

PR #1293, opened 2022-09-22
[fix] - MetaMask Connect Flow

aaronbarnardsound:
### Description
This PR fixes some edge cases that occur when connecting to MetaMask. If a connect request is made and the user clicks outside of the popup menu (which hides the login/connect popup), and then tries to connect again, the connect request will throw an error and Onboard will not be aware that there is another connect request in progress. This gets especially complicated when the `autoSelect` feature is used modals disabled as the app cannot know that the original request is hanging leading to a bad UX as outlined in #1279. This fix will work for any wallets that throw an RPC error for duplicate connect requests. As far as I am aware only MetaMask does this at this time.

The fix does the following:

- Ensures that only one connect request is in flight at any time. It does this via canceling the any old requests whenever a new connect request is fired.
- Ensures that when the user does eventually connect their MetaMask wallet that Onboard will recognize that connection separately from the connection request as that is unreliable in this edge case. It does this by listening to the `accountsChanged` event when an account access already requested RPC error is caught.
- Adds some new copy to the connecting modal to indicate that a pending connect request is awaiting user action and that the user needs to manually open the MetaMask menu as it will not popup a second time.

### Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.8.5-alpha.1",
+  "version": "2.8.5-alpha.2",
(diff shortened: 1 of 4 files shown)


Adamj1232 (approved):
Nice!!

PR #1294, opened 2022-09-22
[chore] - Update homepage links

aaronbarnardsound:
### Description
- Updates homepage link in package.json file for all packages

Note: I did not do a version bump on the packages as they have all been bumped in develop via #1290 so no need to bump again.


Adamj1232 (approved):
Lgtm!!

PR #1295, opened 2022-09-23
docs site update

gesquinca:
### Checklist
- [ ] I have run `yarn check` & `yarn build` to confirm there are not any associated errors


docs/src/app.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%sveltekit.assets%/favicon.png" />
+    <link rel="icon" type="image/x-icon" href="/favicon.png" />
```
taylorjdawson, line 5: @gesquinca I think we may want to keep `%sveltekit.assets%` ?
```suggestion
    <link rel="icon" type="image/x-icon" href="%sveltekit.assets%/favicon.png" />
```

taylorjdawson (changes requested):

docs/src/app.html
gesquinca: From what I can see, it doesn't work correctly with `%sveltekit.assets%/`. That's why I removed it. To try and see if it works correctly without it.

taylorjdawson (approved):

PR #1301, opened 2022-09-28
Enhancement: Sync accounts

### LINKED ISSUE #1225: [Feature]: Make account center request wallet permission if it's not the currently active one
### Is your request related to a problem?

This request is inspired by #1219 

### Feature Description

Account Center widget shows the history of connected metamask wallets even though some of them could have been disconnected already by the user in the metamask directly and since there is no 'wallet disconnected' event we just don't know when this happens. Here's an example 
image
As you can see I only have ...a42d wallet connected in metamask and have since disconnected ihor.eth yet I can select it in the Account Center. 
I propose that whenever the metamask wallet is selected in the account center we send the [wallet_getPermissions](https://docs.metamask.io/guide/rpc-api.html#wallet-getpermissions) request first and look inside the caveats array. 
image
If the wallet is not there anymore we send the [wallet_requestPermissions](https://docs.metamask.io/guide/rpc-api.html#wallet-requestpermissions) request so that the user can select that wallet again and make it active. We can also "clean up" other wallet addresses that are still lingering in the Account Center at this time.

### Alternative Solutions

Not that I'm aware of, I've looked all over Metamask docs.

### Anything else?

We could also send the [wallet_requestPermissions](https://docs.metamask.io/guide/rpc-api.html#wallet-requestpermissions) request whenever someone 
1) clicks on Add Account 
image
2) clicks "Connect another wallet" and selects Metamask from the modal window
image
image

aaronbarnardsound:
### Description
Currently, if a user were to connect multiple MetaMask accounts and then manually disconnect one of those accounts, the disconnected account will still appear connected in the internal `wallets` state and associated UI's driven by that state.

This change implements logic to handle this for wallets that support the `wallet_getPermissions` RPC method, which allows us to check which accounts are currently connected and compare them with our internal connected accounts state and then sync them.

Closes #1225 

### Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.8.5",
+  "version": "2.9.0-alpha.1",
   "description": "Web3-Onboard makes it simple to connect Ethereum hardware and software wallets to your dapp. Features standardized spec compliant web3 providers for all supported wallets, framework agnostic modern javascript UI with code splitting, CSS customization, multi-chain and multi-account support, reactive wallet state subscriptions and real-time transaction state change notifications.",
   "keywords": [
     "Ethereum",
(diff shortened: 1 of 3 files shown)


taylorjdawson (approved):
LGTM!

PR #1307, opened 2022-10-02
Release 2.13.0

aaronbarnardsound:
### ✨ Highlights

- **Enhancement: Sync accounts #1301**

### 📦 Changes per package

- `@web3onboard/core` **v2.9.0**


packages/core/src/provider.ts
```diff
@@ -390,3 +409,41 @@ export function addNewChain(
...
+    return Array.isArray(permissions) ? permissions : []
+  } catch (error) {
+    return []
+  }
+}
+
+export async function syncWalletConnectedAccounts(
+  label: WalletState['label']
+): Promise<void> {
+  const wallet = state.get().wallets.find(wallet => wallet.label === label)
```
Adamj1232, line 430: Do we need a guard to confirm `wallet` exists or are we sure it's there at this step?

Adamj1232 (approved):
Approved with one small question!

packages/core/src/provider.ts
aaronbarnardsound: Wallet should exist since this function is only called within the accounts changed listener which gets removed if a wallet is disconnected. There is also a try catch block around the call to this function for graceful failure.

PR #1327, opened 2022-10-20
Add the Bitski extension injected provider

pzuraq:
### Description

Adds support for the [Bitski](https://www.bitski.com/) browser extension (currently in beta) in the injected provider. 

### Checklist


chronicIntrovert (approved):

packages/injected/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/injected-wallets",
-  "version": "2.2.4",
+  "version": "2.3.0",
```
Adamj1232, line 3:
```suggestion
  "version": "2.3.0-alpha.1",
```
This will allow an alpha package be published for testing purposes

pzuraq: Sounds good, I've updated it accordingly!

Adamj1232:
@pzuraq thanks for the PR! Has this been thoroughly tested with a Bitski wallet on your end?

pzuraq:
Hey @Adamj1232, yes, we have tested it locally using the demo app included in the repository and it appears to be working. You can also test it out yourself by installing the extension here: https://chrome.google.com/webstore/detail/bitski/feejiigddaafeojfddjjlmfkabimkell

We're currently prepping for our open beta and there are some bugs and issues (the activity page is not accurate at the moment, some test data was left in by mistake) but you should be able to test out connecting with it!

Adamj1232:
@pzuraq I am unable to test...after wallet download and account setup the extension is constantly in a loading state - any thoughts?
Screen Shot 2022-10-24 at 10 33 18 AM

pzuraq:
Ah, I think this may be because we made some last minute breaking changes to our graphql schema, we're currently waiting on the Chrome webstore to approve the updated version. I can let you know when that's gone through and you should be able to test then!

Adamj1232:
@pzuraq no problem, it just started working!. I was able to connect but switching chains didn't seem to work as expected. Is that still in development?

Adamj1232 (approved):
Great job, thank you!

Adamj1232:
@pzuraq it would be really cool to see an approach similar to Coinbase's wallet where they have a providers array on the provider to catch any overriden wallets within the window.ethereum space along with the currently injected(Bitski in this case). Its a great way to help share the window.ethereum space and allow the community to grow together rather than competing for it

pzuraq:
@Adamj1232 yes absolutely, we're planning on figuring that out during the beta period, we were just trying to figure out the best/most compatible way to do it. We saw this EIP which we think would be ideal: https://eips.ethereum.org/EIPS/eip-5749, but that's obviously still a draft so we aren't implementing it yet. 

Hopefully that advances soon, but in the meantime is the best thing to do to mimic Coinbase?

Adamj1232:
@pzuraq ahh great to see an EIP forming around this!
Great to hear!
Out of curiosity why are you setting the `isMetamask` flag to true?
With CB's market coverage it would be a solid choice to mimic what they are doing. 
Because the space is already `window.ethereum` calling it `evmproviders` seems redundant but we will support whatever naming convention that passes along with any large coverage wallet(Coinbase)

pzuraq:
Ah yes, that's something we were doing when testing just to make sure the extension worked with dapps before launch, we're planning on removing shortly (thanks for reminding me!)

Will do re `window.ethereum.providers`, that sounds like a good solution until the EIP moves forward.

PR #1333, opened 2022-10-24
Docs gas example

Adamj1232:
### Description
Add more examples to the Gas modules along with a gas comparison visualization to compare against RPC gas pricing

Screen Shot 2022-10-24 at 1 07 49 PM

### Checklist

### DIFF
--- a/docs/src/lib/components/gas/Gas.svelte
+++ b/docs/src/lib/components/gas/Gas.svelte
@@ -5,7 +5,7 @@
   import gasModule from '@web3-onboard/gas'
   import { onMount } from 'svelte'
   import { ethers } from 'ethers'
-  import type { GasPrice, RPCGasPrice } from './types'
+  import type { GasPrice, RPCGasPrice, GasData } from './types'
 
   let cardBg: any
   let animation: any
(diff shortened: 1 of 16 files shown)


docs/tailwind.config.cjs
```diff
@@ -11,7 +11,20 @@ module.exports = {
...
+        conf: {
+          99: '#5aea98',
+          95: '#5dea5a',
+          90: '#bcea5a',
+          80: '#ffe600',
+          70: '#eab05a'
+        }
+      }
+    }
+  }
```
gesquinca, line 27: @Adamj1232 What are these colors for?

Adamj1232: Those colors are for the gradient on the gas page

gesquinca: Can we scope them down to just the gas component? If we're only using them there, I'd rather not add them to global theme.

docs/src/lib/components/gas/Gas.svelte
```diff
@@ -0,0 +1,84 @@
...
+    estimatedTransactionCount: null,
+    seconds: null
+  }
+</script>
+
+<div class="Gas px-6  p-4">
+  <div class="flex whitespace-nowrap mb-3 text-sm select-none">
+    <span class="flex items-center">MORE LIKELY</span>
+    <span
+      class="bg-gradient-to-r from-conf-99 via-conf-95 via-conf-90 via-conf-80 to-conf-70 h-[1px] mx-2 my-3 w-full rounded-full"
```
Adamj1232, line 71: @gesquinca Usage of gradient colors

docs/tailwind.config.cjs
Adamj1232: @gesquinca they are pretty specific variables and would be required to be he to use the tailwinds implementation of gradients. This pattern follows what we are doing in other projects(Gas extensions). I tagged you in the implementation.

Adamj1232: All cleaned up!

gesquinca (approved):

PR #1347, opened 2022-10-31
Enable new domain at Gnosis package

dasanra:
### Description
In the process of the Safe rebranding we will introduce several changes. Among others, during this week the app will be deployed on a new domain. With the current default configuration web3-onboard doesn't allow to integrate dApps as Safe Apps in the new domain. We propose to add it to the default configuration.

### Checklist


dasanra:
@Adamj1232 hey yes, I was also checking that other ticket. It's likely that we will need to update the Safe Apps SDK to add some changes there. We were planning to do it on a separate PR because the latest version of the SDK contains breaking changes compared to the one you are using, and we will need to deliver a major version of this package too, so we would like to handle that with less pressure.

We would also like to suggest some other changes on that PR upgrading the SDK version, but this small change is kind of urgent for us, as some dApps developers are waiting for it to not loose compatibility with the Safe.

The app is already deployed on the new domain, so you can also check it there

Adamj1232 (approved):

PR #1359, opened 2022-11-11
[transaction-preview: v2.0.0-alpha.1 , core: v2.11.0-alpha.1, react, vue]: Feature - Transaction Preview Package

Adamj1232:
### Description
This PR creates the new Transaction Preview package.

These demos use mock transaction code to create a contract interaction swapping dai for eth
Live demo w/ default settings - https://636eb637075a5f423cd40deb--sage-moxie-d48c11.netlify.app/
Live demo with `requireTransactionApproval` prop set to true - https://heroic-crostata-41f374.netlify.app/

Views:
Latest design updates:
preview4
previewSS4

TP with AC & Notify enabled positioned upper right through W3O config
Screen Shot 2022-11-09 at 1 14 33 PM

TP with AC & Notify enabled positioned lowerLeft through W3O config
Screen Shot 2022-11-09 at 1 15 10 PM

TP without AC w/ Notify enabled positioned in upper right through W3O config
Screen Shot 2022-11-09 at 1 05 27 PM

TP with AC enabled in lower left & w/ Notify enabled positioned in upper right through W3O config
Screen Shot 2022-11-09 at 1 14 01 PM

TP without AC w/ Notify enabled positioned in lower left through W3O config
Screen Shot 2022-11-09 at 1 12 59 PM

TP without AC or Notify enabled positioned default(upper right) there is no config to change just TP position
Screen Shot 2022-11-09 at 1 12 24 PM

Mobile - TP with AC & Notify enabled positioned in upper through W3O config
Screen Shot 2022-11-09 at 1 52 23 PM

TP with requireTransactionApproval enabled
Screen Shot 2022-11-11 at 1 31 59 PM

API: 
```typescript

export type TransactionPreviewModule = (
  options: TransactionPreviewInitOptions
) => TransactionPreviewAPI

export type TransactionPreviewAPI = {
  /**
   * Pass this method a standard EIP1193 provider
   * (such as an injected wallet from window.ethereum)
   * and it will be patched to allow for transaction previewing
   */
  patchProvider: (provider: PatchedEIP1193Provider) => PatchedEIP1193Provider
  /**
   * Pass this method any full, presigned ethereum transaction
   * for previewing and simulation
   */
  simTransactions: (txs: [TransactionObject]) => Promise<SimPlatformResponse>
  /**
   * Pass this method an HTML element ID to allow for
   * the Transaction Preview UI to mount to it.
   * Note: The element must exist within the DOM tree
   *  at time of preview/rendering
   */
  setContainerElement: (elementId: string) => void
  /**
   * This property will return the container element HTML ID
   *  set for the Transaction Preview UI to mount to
   */
  containerElement?: string
}

export type PatchedEIP1193Provider = EIP1193Provider & { simPatched: boolean }

export interface ProviderReq {
  method: string
  params?: Array<unknown>
}

export type RequestOptions = Pick<TransactionPreviewInitOptions, 'apiKey'>

export type TransactionPreviewInitOptions = {
  /**
   * Blocknative API key (https://explorer.blocknative.com/account)
   */
  apiKey: string
  /**
   * Your Blocknative API secret key: Add a Secret Key to your API key
   * by using the three dot menu next to the name of your API key.
   * (https://docs.blocknative.com/account#secret-key)
   * */
  secretKey: string
  /**
   * Optional dom query string to mount UI to
   * */
  containerElement?: string
  /**
   * Optional requirement for user to accept transaction balance changes
   * prior to sending the transaction to the wallet
   * */
  requireTransactionApproval?: boolean
  /**
   * An optional internationalization object that defines the display
   * text for different locales. Can also be used to override the default text.
   *  To override the default text, pass in a object for the en locale
   */
  i18n?: i18nOptions
}
```

### Usage with Web3-Onboard Core package (recommended)

```typescript
import Onboard from '@web3-onboard/core'
import injectedModule from '@web3-onboard/injected'
import transactionPreviewModule from '@web3-onboard/transaction-preview'

const injected = injectedModule({})
const transactionPreview = transactionPreviewModule({
  apiKey: 'xxxxx-4729-457a-8d76-d6d15692657b',
  secretKey: 'xxxxx-f0e8-40ec-b49a-2571715642cb'
})

const onboard = Onboard({
  transactionPreview,
  wallets: [injected],
  chains: [
    {
      id: '0x1',
      token: 'ETH',
      label: 'Ethereum',
      rpcUrl: 'https://mainnet.infura.io/v3/17c1e1500e384acfb6a72c5d2e67742e'
    }
  ]
  // ... other Onboard options
})

// Transaction code here using Ether.js or Web3.js or custom
// The transaction will automatically be picked up and simulated with a UI displaying in the upper right corner
```

### Standalone Usage

If you would like to use outside of web3-onboard wallet connect more steps are required

```typescript
import transactionPreviewModule from '@web3-onboard/transaction-preview'
const transactionPreview = transactionPreviewModule({
  apiKey: 'xxxxx-4729-457a-8d76-d6d15692657b',
  secretKey: 'xxxxx-f0e8-40ec-b49a-2571715642cb',
  // Require balance change confirmation prior to sending to wallet
  requireTransactionApproval: true
})

transactionPreview.setContainerElement(
  '#existing-html-element-to-append-preview-ui'
)

// Can be any standardized EIP1193 Wallet Provider
transactionPreview.patchProvider(window.ethereum)

// Transaction code here using Ether.js or Web3.js or custom

// UI will display within DOM element specified as containerElement
```

Alternatively Transaction Preview can be ran headless without a UI

```typescript
import transactionPreviewModule from '@web3-onboard/transaction-preview'
const transactionPreview = transactionPreviewModule({
  apiKey: 'xxxxx-4729-457a-8d76-d6d15692657b',
  secretKey: 'xxxxx-f0e8-40ec-b49a-2571715642cb'
})

// Transaction code here using Ether.js or Web3.js or custom
// Pass unsigned transactions into the below method to receive a transaction preview for each transaction passed

transactionPreview.simTransactions(unsignedTransaction)
```

### Checklist


packages/transaction-preview/src/index.ts
```diff
@@ -0,0 +1,259 @@
...
+      req.method === 'eth_sendTransaction' &&
+      req.params &&
+      req.params.length
+    ) {
+      const transactionParams =
+        req.params as EthSignTransactionRequest['params']
+      try {
+        const preview = await simulateTransactions(options, transactionParams)
+        if (preview.status !== 'simulated') {
+          // If transaction simulation was unsuccessful do not create DOM el
```
aaronbarnardsound, line 108: Do we need to proxy the request to the provider.request here regardless of simulation result?

packages/transaction-preview/src/index.ts
```diff
@@ -0,0 +1,259 @@
...
+          return
+        }
+        if (app) app.$destroy()
+        app = mountTransactionPreview(preview)
+        options.requireTransactionApproval
+          ? handleRequireApproval(app, fullProviderRequest, req)
+          : fullProviderRequest(req).then((hash: unknown) => {
+              hash && app.$destroy()
```
aaronbarnardsound, line 116: Do we want to destroy the Sim UI regardless of if there is a `hash` here?

packages/transaction-preview/src/index.ts
```diff
@@ -0,0 +1,259 @@
...
+            }).catch(() => app.$destroy())
+      } catch (e) {
+        throw new ProviderRpcError({
+          code: ProviderRpcErrorCode.ACCOUNT_ACCESS_REJECTED,
```
aaronbarnardsound, line 120: Is this the correct code for transaction rejected?

packages/transaction-preview/src/index.ts
```diff
@@ -0,0 +1,259 @@
...
+        }
+        if (app) app.$destroy()
+        app = mountTransactionPreview(preview)
+        options.requireTransactionApproval
+          ? handleRequireApproval(app, fullProviderRequest, req)
+          : fullProviderRequest(req).then((hash: unknown) => {
+              hash && app.$destroy()
+            }).catch(() => app.$destroy())
+      } catch (e) {
+        throw new ProviderRpcError({
```
aaronbarnardsound, line 119: Do we want to destroy the Sim UI here in the case of failure?

packages/transaction-preview/src/index.ts
```diff
@@ -0,0 +1,259 @@
...
+          code: ProviderRpcErrorCode.ACCOUNT_ACCESS_REJECTED,
+          message: `User rejected the transaction: ${e}`
+        })
+      }
+    } else {
+      return fullProviderRequest(req)
+    }
+  }
+  patchedProvider.request = request
+  patchedProvider.simPatched = true
```
aaronbarnardsound, line 129: I wonder if we need to wrap the whole call in a try catch statement? I vaguely remember that some providers are set to read only, so we may have trouble with some wallets when setting the `simPatched` value, but I could be wrong on that. We could log a warning to the console in the case that we could not patch the provider.

packages/transaction-preview/src/index.ts
```diff
@@ -0,0 +1,259 @@
...
+  return patchedProvider
+}
+
+const transactionPreview: TransactionPreviewModule = (
+  initOptions: TransactionPreviewInitOptions
+): TransactionPreviewAPI => {
+  const { i18n, containerElement } = initOptions
+  if (initOptions) {
+    const error = validateTPInit(initOptions)
```
aaronbarnardsound, line 138: It might be worth validating the options before destructuring from them?

packages/transaction-preview/src/simulateTransactions.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+    return cleanedTrans
+  })
+
+  const body = {
+    system: 'ethereum',
+    network: 'main',
+    transactions: cleanedTransactions
+  }
+  console.log(body)
```
aaronbarnardsound, line 34: Sneaky log

packages/transaction-preview/src/views/Index.svelte
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const transactionApproved = (approved: boolean) => {
+    approved$.next(approved)
+  }
+  let mounted = false
+  onMount(() => (mounted = true))
+</script>
+
+{#if mounted}
```
aaronbarnardsound, line 32: Curious why this conditional was needed? Svelte should handle all of this internally. Or is this to ensure animation on first mount? If so you can add the [`intro` option](https://svelte.dev/docs#run-time-client-side-component-api) when creating the component.

packages/transaction-preview/src/views/Maximized.svelte
```diff
@@ -0,0 +1,371 @@
...
+  export let requireTransactionApproval: boolean
+  export let transactionApproved: (approved: boolean) => void
+  export let toggleExpanded: (maximize: boolean) => void
+  export let destroyApp: () => void
+  export let simResponse: SimPlatformResponse
+  export let startTime: number
+  const device = getDevice()
+
+  let transactionOriginator = simResponse.transactions[0].from
+  let balanceChanges = simResponse.netBalanceChanges.reduce((arr, changes) => {
```
aaronbarnardsound, line 20: Can these be `const` declarations since they don't seem to be reassigned?

packages/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/common",
-  "version": "2.2.3",
+  "version": "2.2.4-alpha.1",
```
aaronbarnardsound, line 3: Are there changes in the common package that require a version bump?

packages/injected/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/injected-wallets",
-  "version": "2.3.0",
+  "version": "2.3.1-alpha.1",
```
aaronbarnardsound, line 3: Are there changes in the injected package that require a version bump? I see the bump for the common package, but there are no changes I can see in the common package.

aaronbarnardsound (commented):
Awesome work! A few questions and comments in there, but looks pretty good.

packages/transaction-preview/src/views/Index.svelte
Adamj1232: Yep that was the reasoning, thank you for this! Way cleaner!

packages/transaction-preview/src/index.ts
Adamj1232: Good call! I will add a destroy the UI if it exists

packages/transaction-preview/src/index.ts
Adamj1232: Oh yeah, good call no need to check for the hash

packages/transaction-preview/src/index.ts
Adamj1232: @aaronbarnardsound Thats a good question! When a transaction is rejected, atleast through MetaMask the code is 4001 the same as `ACCOUNT_ACCESS_REJECTED`
I could create a new enum with the same value for this situation within the common package as this is the second time this code in this context has come up and would add clarity. Thoughts?
Screen Shot 2022-11-14 at 3 48 52 PM

packages/transaction-preview/src/index.ts
Adamj1232: Ah good call, I think Binance wallet is one of those!

Adamj1232: @aaronbarnardsound I added a try around the reassigning of the request property.
```  
try {
    patchedProvider.request = request
    patchedProvider.simPatched = true
  } catch (err) {
    console.error(
      `There was an error patching the passed in wallet provider. 
      The provider may be read only and may be incompatible with Transaction Preview`
    )
  }
  ```
  Is that along the lines of what you had in mind?

packages/transaction-preview/src/index.ts
Adamj1232: Ah yep totally, if not simulated the request should pass right on through to the original request handler

packages/transaction-preview/src/index.ts
aaronbarnardsound: No worries, just double checking. No need to add to the enum I don't think.

packages/transaction-preview/src/index.ts
aaronbarnardsound: Yep!

aaronbarnardsound (approved):
Looks good 🚀

taylorjdawson (approved):
Approved! with 1 comment

gesquinca (approved):

PR #1371, opened 2022-11-17
[BUG] fix coinbase default entry

duxiaofeng-github:
### Description
https://github.com/wagmi-dev/wagmi/issues/383, as discussed here, the coinbase wallet export constructor as default, and vite(or not caused by it) will wrap by another default. so we need to detect if the default export is the constructor. if not, use the next default.

image

### Checklist

### DIFF
--- a/packages/coinbase/src/index.ts
+++ b/packages/coinbase/src/index.ts
@@ -13,12 +13,17 @@ function coinbaseWallet({
         const [chain] = chains
         const { name, icon } = appMetadata || {}
 
-        const { CoinbaseWalletSDK } = await import('@coinbase/wallet-sdk')
+        const { default: CoinbaseWalletSDK } = await import(
+          '@coinbase/wallet-sdk'
+        )
+        const CoinbaseWalletSDKConstructor = (CoinbaseWalletSDK as any).default
+          ? (CoinbaseWalletSDK as any).default
+          : CoinbaseWalletSDK
 
         const base64 = window.btoa(icon || '')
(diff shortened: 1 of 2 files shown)


duxiaofeng-github:
@Adamj1232 fixed

packages/coinbase/src/index.ts
```diff
@@ -13,12 +13,22 @@ function coinbaseWallet({
...
 
-        const { CoinbaseWalletSDK } = await import('@coinbase/wallet-sdk')
+        // according to https://github.com/wagmi-dev/wagmi/issues/383
+        // @coinbase/wallet-sdk export double default fields
+        // so we need to detect it to get the real constructor
+        const { default: CoinbaseWalletSDK } = await import(
+          '@coinbase/wallet-sdk'
+        )
+        const CoinbaseWalletSDKConstructor = (
+          (CoinbaseWalletSDK as any).default
```
taylorjdawson, line 23: @Adamj1232 if you check for `'default' in` I think you could forgo all of the casting ya?

Adamj1232: @duxiaofeng-github ☝🏼

Adamj1232: Would still get a type error getting trying to access `.default` after checking using `in`

Adamj1232, line 23:
```suggestion
          'default' in CoinbaseWalletSDK
```

Adamj1232 (approved):
Thank you again!

PR #1383, opened 2022-11-29
Docs release 

gesquinca:
### Description
Publish latest updates to documentation site


Adamj1232 (approved):
Looks good! Not sure we need the common module in there since it isn't installed separately

docs/src/routes/docs/[...3]modules/common.md
```diff
@@ -0,0 +1,22 @@
+# Common
```
Comment (Adamj1232) [inline docs/src/routes/docs/[...3]modules/common.md:1]: Do we need this? No one should ever have to install this separately

Comment (gesquinca) [reply docs/src/routes/docs/[...3]modules/common.md:1]: Good catch. We'll remove.

PR #1410, opened 2022-12-08
Add Unstoppable Domains Reverse Resolution

V-Shadbolt:
### Description
Added reverse resolution of wallet addresses into an unstoppable domain within the core package. Resolves whether the user connects with Unstoppable or not.

### Checklist

###Screenshots
Screenshot 2022-12-07 at 4 37 29 PM
Screenshot 2022-12-07 at 4 38 14 PM


packages/core/rollup.config.js
```diff
@@ -35,6 +36,9 @@ export default {
       sourceMap: !production,
       inlineSources: !production
     }),
+    commonjs({
+      include: /node_modules/
```
Adamj1232, line 40: @V-Shadbolt quick question here, can you provide a little background on the reason for this?
Also, can it be scoped to the ustoppabledomains package?
Thanks!

V-Shadbolt: @Adamj1232 Yup! This was added because rollup wasn't including the unstoppable domains node module when compiling which threw a bunch of errors when the functions were called.  

The commonjs plugin instructs rollup to convert node modules to ES6 so that they can be interpreted by browsers and the `include: /node_modules/` flag is for the directory of the modules to include during compilation 

I don't believe this can be scoped back unless we can specify just the unstoppable module with something like this:  `include: /node_modules/@unstoppabledomains/`.

Adamj1232: Ah gotcha, thank you for the fast response! `include: /node_modules/@unstoppabledomains/` is what I had in mind.
I can open a PR or you can but want to get your teams approval on the changes 🤓

V-Shadbolt: Hey! Managed to get it rescoped to just Unstoppable using the below snippet while retaining the original functionality of this PR. 

`
commonjs({
      namedExports: {
        '../../node_modules/@unstoppabledomains/resolution/build/index.js': [ 'Resolution' ]
      },
}),`

PR #1426, opened 2022-12-14
Release: 2.16.0

taylorjdawson:


packages/demo/webpack.config.js
```diff
@@ -8,6 +8,9 @@ const prod = mode === 'production'
 module.exports = {
   entry: './src/main.js',
   resolve: {
+    fallback: {
```
Adamj1232, line 11: @taylorjdawson Can we add this to the webpack config in the Core Readme?

packages/core/src/utils.ts
```diff
@@ -12,22 +12,23 @@ import type {
...
+  polygonIcon,
+  binanceIcon,
+  questionIcon,
+  fantomIcon,
+  optimismIcon,
+  celoIcon,
+  avalancheIcon,
+  harmonyOneIcon,
+  arbitrumIcon
+} from './icons'
```
aaronbarnardsound, line 31: Does this need to be `'./icons/index.js'` to maintain ESM compatibility?

packages/core/src/views/account-center/Maximized.svelte
```diff
@@ -10,10 +10,8 @@
   import connect from '../../connect.js'
   import disconnect from '../../disconnect.js'
   import { state } from '../../store/index.js'
-  import WalletAppBadge from '../shared/WalletAppBadge.svelte'
   import { getDefaultChainStyles, unrecognizedChainStyle } from '../../utils.js'
-  import SuccessStatusIcon from '../shared/SuccessStatusIcon.svelte'
-  import NetworkSelector from '../shared/NetworkSelector.svelte'
+  import { NetworkSelector, SuccessStatusIcon, WalletAppBadge } from '../shared'
```
aaronbarnardsound, line 14: Same with all of these imports, I think we need the full extension to maintain compatibility with ESM standard.

packages/demo/webpack.config.js
taylorjdawson: Good idea!

Adamj1232 (approved):
Good work man!! Fire it up 🔥 🚀  LGTM!

PR #1452, opened 2023-01-09
[core-v2.13.0-alpha.2, react, vue]: Enhancement - Container el prop for Connect Modal

Adamj1232:
### Description
This PR adds the ability to pass an html container el ID through the W3O init and append the connect modal to that element

Per a slack convo with Gustavo mobile styles will not be changed for this updates as it is out of scope.

```typescript
export type ContainerElements = {
  accountCenter?: string
  connectModal?: string
}
```

If both elements have a containerEl defined each will be appended to the defined container element as seen in img #2

Screen Shot 2023-01-09 at 1 34 26 PM

Screen Shot 2023-01-09 at 1 48 18 PM

### Checklist


aaronbarnardsound (approved):

omarsy:
@Adamj1232  Maybe we need also change the behaviour for Hardware wallet. I think we always have modal even if we choose inline

Adamj1232:
> @Adamj1232 Maybe we need also change the behaviour for Hardware wallet. I think we always have modal even if we choose inline

Ah yep we will need to make a similar update for the HW-common package - as this is a separate package it will take some time. As for injected/sdk wallets does this change meet your needs?

omarsy:
@Adamj1232 Yes Thank you I will test it. 🔥
On my side I will put in the same container maybe we can make the container of hw overridable, what do you think ?

Adamj1232:
@omarsy with the update I'm working on the styling/overlay of the HW Account Select modal over the connect modal will be up to the dev to implement. If the same container id is passed to the hw wallet init and containerElements.connectModal they will display next to each other. So it will be up to the dev to control how those are shown using the container elements styling. Just a heads up 😄

omarsy:
@Adamj1232  Thank You.
I see we always have the modal icon and the css related to the modal can we delete it when we are in inline mode
image

Adamj1232:
@omarsy Ah you are referring to the close button/icon on the connect modal? We can add a prop to disable this ability if needed

PR #1474, opened 2023-01-18
Fix: Torus Chain Changed Error

### LINKED ISSUE #1446: [Bug]: Switching chain breaks when connected through Torus module
### Current Behavior

When connected with a Torus wallet after initial connection and wanting to switch chains, after confirming the switch in wallet the app gets stuck on switching status (see screens)

### Expected Behavior

Should be able to switch chains

### Steps To Reproduce

1. reproducible on https://reactdemo.blocknative.com/
2. locally I ran setChain hook while connected with torus

### What package is effected by this issue?

@web3-onboard/torus

### Is this a build or a runtime issue?

Runtime

### Package Version

2.12.1

### Node Version

16

_No response_

### Relevant log output

_No response_

### Anything else?

(local build)

(react-demo)

### Sanity Check

- [X] If this is a build issue, I have included my build config. If this is a runtime issue, I have included reproduction steps and/or a [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example).

aaronbarnardsound:
### Description
The Torus provider `chainChanged` event now returns a hex encoded string rather than decimal. The current patching logic does not handle hex encoded `chainId`. There is no need for the patch now, so this PR removes it.

Fixes #1446 

### Checklist

### DIFF
--- a/packages/demo/package.json
+++ b/packages/demo/package.json
@@ -39,7 +39,7 @@
     "@web3-onboard/phantom": "^2.0.0-alpha.1",
     "@web3-onboard/portis": "^2.1.3",
     "@web3-onboard/sequence": "^2.0.4",
-    "@web3-onboard/torus": "^2.2.0",
+    "@web3-onboard/torus": "^2.2.1-alpha.1",
     "@web3-onboard/trezor": "^2.3.2",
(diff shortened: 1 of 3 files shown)


Adamj1232 (approved):
Nice!!

PR #1500, opened 2023-01-30
feat: WalletConnect v2 connector

ganchoradkov:
### Description
Implemented WalletConnect v2 provider similar to your existing v1 provider.Also added an extra optional property to `WalletConnectOptions` called `version` that specifies which protocol to load (v1 or v2). This update should make migrating to v2 as seamless as
- set `version : '2'`
- set `projectId` - required key that can be obtained [here](https://cloud.walletconnect.com) 

To avoid unintended breaking changes, `v1` is loaded by default. When v1 is shutdown (scheduled March 2023), small update can be done to load v2 by default.
### Checklist

- [x] send transaction
- [x] switch chains
- [x] sign message
- [x] sign typed message
- [x] disconnect

- [x] send transaction
- [x] switch chains
- [x] sign message
- [x] sign typed message
- [x] disconnect


Adamj1232 (changes requested):
Can you update the README for the new options and usage example for both V1 &  V2?

Are all the other init options still compatible with WC v2?

Also wondering if there was a reason around using strings for the version number?

Looks good will start some testing today!

I updated the yarn lock to fix the build errors and bumped the package version.
If you can give a pull when you have a moment that would be great.

Adamj1232 (approved):
Thank you a ton for the collaboration! Really enjoyed working with the WalletConnect team!

PR #1506, opened 2023-02-01
Feature: Auto Connect Last Wallet Option

aaronbarnardsound:
### Description
This PR implements a commonly requested feature for auto connecting a wallet that has been previously connected. There is now a `autoConnectLastWallet` option which can be set during initialization:

```javascript
const onboard = Onboard({
  // ... other options
  connect: {
    autoConnectLastWallet: true
  }
})
```

Also implemented in this PR is error handling for accessing `localStorage` to handle when browsers have blocked access (eg TOR browser).

### Checklist

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -129,12 +129,16 @@ An object that allows for customization of the Connect Modal and accepts the typ
 ```typescript
 type ConnectModalOptions = {
   showSidebar?: boolean
-  showSidebar?: boolean
   /**
    * Disabled close of the connect modal with background click and
    * hides the close button forcing an action from the connect modal
    */
   disableClose?: boolean // defaults to false
+  /**If set to true, the last connected wallet will store in local storage.
+   * Then on init, onboard will try to reconnect to that wallet with
+   * no modals displayed
+   */
+  autoConnectLastWallet?: boolean // defaults to false
 }
 ```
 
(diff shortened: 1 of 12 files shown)


packages/react/package.json
```diff
@@ -62,7 +62,7 @@
     "typescript": "^4.5.5"
   },
   "dependencies": {
-    "@web3-onboard/core": "^2.14.1",
+    "@web3-onboard/core": "^2.15.0-alpha.1",
```
Adamj1232, line 65: bump the react package versions too?

packages/vue/package.json
```diff
@@ -63,7 +63,7 @@
     "@vueuse/core": "^8.4.2",
     "@vueuse/rxjs": "^8.2.0",
     "@web3-onboard/common": "^2.2.3",
-    "@web3-onboard/core": "^2.14.1",
+    "@web3-onboard/core": "^2.15.0-alpha.1",
```
Adamj1232, line 66: bump the vue package versions too?

packages/core/README.md
```diff
@@ -128,13 +128,17 @@ An object that allows for customization of the Connect Modal and accepts the typ
...
 ```typescript
 type ConnectModalOptions = {
-  showSidebar?: boolean
   showSidebar?: boolean
   /**
    * Disabled close of the connect modal with background click and
    * hides the close button forcing an action from the connect modal
    */
   disableClose?: boolean // defaults to false
+  /**If set to true, the last connected wallet will store in local storage.
```
Adamj1232, line 137: Sweet! We could add these updates to the docs site core module too

packages/core/src/constants.ts
```diff
@@ -31,7 +31,8 @@ export const APP_INITIAL_STATE: AppState = {
 }
 
 export const STORAGE_KEYS = {
-  TERMS_AGREEMENT: 'onboard.js:agreement'
+  TERMS_AGREEMENT: 'onboard.js:agreement',
+  LAST_CONNECTED_WALLET: 'onboard.js:last_connected_wallet'
```
Adamj1232, line 35: Would we want to update the ls key to be prefixed with `web3-onboard` instead of `onboard.js` for clarity?

Adamj1232 (approved):
Few notes but overall looks good!

packages/core/src/constants.ts
aaronbarnardsound: We could, but then all users would need to re-agree to the terms and conditions....

packages/core/README.md
aaronbarnardsound: Yep docs update will come next :)

packages/react/package.json
aaronbarnardsound: Ah yep good catch!

PR #1518, opened 2023-02-08
Feature : Transaction Preview - Add Gas Used to preview view

Adamj1232:
### Description

This PR includes:

- Update to bnc-sdk version to include full simulation types
- Adding of Gas Used table to the Preview view
- Refactor of TP API to add standalone simulation usage
- Update documentation around the new package usage

Screen Shot 2023-02-07 at 3 26 02 PM

### Checklist


docs/src/routes/docs/[...3]modules/transaction-preview.md
```diff
@@ -73,6 +73,102 @@ const onboard = Onboard({
...
+      ...popTransaction,
+      from: addressFrom,
+      to: uniswapV2Router,
+      value: 0
+    }
+  ]
+  await previewTransaction(transactions)
+}
+
+simulate(ethereumProvider)
```
Comment (aaronbarnardsound) [inline docs/src/routes/docs/[...3]modules/transaction-preview.md:169]: Does the `previewTransaction` function return the simulated payload? If so, i reckon the example should show a response being returned ie: `const simulated = await simulate(ethereumProvider)`

packages/transaction-preview/README.md
```diff
@@ -52,6 +53,102 @@ const onboard = Onboard({
 // The transaction will automatically be picked up and simulated with a UI displaying in the upper right corner
 ```
 
+### Standalone Usage
+
+To use the Transaction Preview package with web3-onboard all a developer needs to do is initialize web3-onboard with their [Blocknative API key](https://onboard.blocknative.com/docs/overview/introduction#optional-use-an-api-key-to-fetch-real-time-transaction-data-balances-gas) and pass in the module as shown below.
```
aaronbarnardsound, line 58: I found this section a little confusing as the heading is about standalone usage, but the paragraph describes how to use with Onboard.

aaronbarnardsound (approved):
Looks good. Just a couple of comments around docs

gesquinca (approved):
I looked at the code changes, but could not test locally.

docs/src/routes/docs/[...3]modules/transaction-preview.md
```diff
@@ -85,12 +181,24 @@ export type TransactionPreviewAPI = {
...
+  
   /**
    * Pass this method a standard EIP1193 provider
    * (such as an injected wallet from window.ethereum)
    * and it will be patched to allow for transaction previewing
    */
   init: (initializationOptions: TransactionPreviewInitOptions) => void
+  
+  /**
+   * Pass this method a transaction that would be passed to a wallet provider
```
Comment (taylorjdawson) [inline docs/src/routes/docs/[...3]modules/transaction-preview.md:193]: This method accepts a transaction meant for a wallet provider (created using libraries like Ethers or Web3), simulates the transaction and generates a corresponding UI.

@Adamj1232 maybe better way to word it??

Comment (Adamj1232) [reply docs/src/routes/docs/[...3]modules/transaction-preview.md:193]: Yep I like that a lot better, gracias!!

Comment (Adamj1232) [reply docs/src/routes/docs/[...3]modules/transaction-preview.md:193]: This was super helpful and helped me refine the other two as well. TS notes FTW!

docs/src/routes/docs/[...3]modules/transaction-preview.md
Comment (Adamj1232) [reply docs/src/routes/docs/[...3]modules/transaction-preview.md:168]: It does not return the payload it will render the TP modal to the containerEl passed into the init - do you think it would be valuable to return the whole payload as well?
I had a similar thought but was kind of on the fence around that

packages/transaction-preview/README.md
Adamj1232: Ah good catch!

docs/src/routes/docs/[...3]modules/transaction-preview.md
Comment (Adamj1232) [reply docs/src/routes/docs/[...3]modules/transaction-preview.md]: Added `MultiSimOutput` response to `previewTransaction` function and updated this example....thanks guys!

packages/transaction-preview/src/views/Maximized.svelte
```diff
@@ -42,7 +42,14 @@
 
   const cleanBalance = (dirtyBalance: string): string => {
     const gweiToEther = ethers.utils.formatEther(dirtyBalance)
-    const roundTo4Decimal = parseFloat(gweiToEther).toFixed(4)
+    const roundTo4Decimal = parseFloat(gweiToEther).toFixed(6)
```
taylorjdawson, line 45: Don't put the number of decimals IMO as then we get issues like this where it was changed from `4` to `6` and now you have to update the variable names every time precision changes 
```suggestion
    const roundGwei = parseFloat(gweiToEther).toFixed(6)
```
@Adamj1232

Adamj1232: Artifact of days of ol....Good catch thank you!

docs/src/routes/docs/[...3]modules/transaction-preview.md
```diff
@@ -73,24 +72,146 @@ const onboard = Onboard({
...
+  )
+  const erc20_contract = new ethers.Contract(oneInch, erc20_interface)
+  const oneEther = ethers.BigNumber.from('9000000000000000000')
+  approveTxData = await erc20_contract.populateTransaction.approve(
+    CONTRACT_ADDRESS,
+    oneEther
+  )
+
+  const amountOutMin = 0
+  const amountOutMinHex = ethers.BigNumber.from(amountOutMin.toString())._hex
```
Comment (taylorjdawson) [inline docs/src/routes/docs/[...3]modules/transaction-preview.md:143]: @Adamj1232 would be better? I would avoid using `_` pre-fixed properties as they are generally considered private and subject to change AFAIK 
```suggestion
  const amountOutMinHex = ethers.BigNumber.from(amountOutMin).toHexString()
```
I think you can remove `toString()` as `BigNumber.from()` accepts both (just tested this)

You could also do this:
```
const amountOutMinHex = '0x0'
```

packages/transaction-preview/README.md
```diff
@@ -52,26 +55,145 @@ const onboard = Onboard({
...
+  )
+  const erc20_contract = new ethers.Contract(oneInch, erc20_interface)
+  const oneEther = ethers.BigNumber.from('9000000000000000000')
+  approveTxData = await erc20_contract.populateTransaction.approve(
+    CONTRACT_ADDRESS,
+    oneEther
+  )
+
+  const amountOutMin = 0
+  const amountOutMinHex = ethers.BigNumber.from(amountOutMin.toString())._hex
```
taylorjdawson, line 126: same here (see [previous comment](https://github.com/blocknative/web3-onboard/pull/1518/files#r1100802575))

packages/transaction-preview/src/index.ts
```diff
@@ -137,6 +111,95 @@ export const patchProvider = (
...
+): Promise<void | unknown> => {
+  try {
+    if (!options) {
+      throw new Error(
+        `Please initialize Transaction Preview package prior to previewing a transaction.
+         You can do this by calling the init function with the appropriate params`
+      )
+    }
+    const preview = await simulateTransactions(options, transaction)
+    if (preview.error.length) {
```
taylorjdawson, line 130: @Adamj1232 is there a case where `preview` | `error` could be undefined here?

packages/transaction-preview/src/index.ts
```diff
@@ -137,6 +111,95 @@ export const patchProvider = (
...
+    (acc: string[], tx: InternalTransaction[]) => {
+      if (tx.length) {
+        tx.forEach((t: InternalTransaction) => {
+          if (t.errorReason) acc.push(t.errorReason)
+        })
+      }
+      return acc
+    },
+    []
+  )
```
taylorjdawson, line 194: because we like to golf ⛳ 
```suggestion
let internalErrs = preview.internalTransactions.reduce(
(acc: string[], tx: InternalTransaction[]) => 
  tx.length ? [...tx.filter(txn => !!txn.errorReason), ...acc] : acc,
[] 
)
```

taylorjdawson: lol probs would want to test but I saw this simplification and couldn't halp myself

packages/transaction-preview/src/views/Maximized.svelte
taylorjdawson: haha figured! lol apologies for the tone I just re-read that lmao
added some fun flair 😝

packages/transaction-preview/src/index.ts
Adamj1232: Lol its shorter but I would say more readable. I get what its doing but how long would it take someone new to this area to read through and try to figure it out? If its even a few seconds/mins more I always go for readability because of maintainability...dont even need to test it to know what its doing, but I do like the clever stuff 🤓

packages/transaction-preview/src/index.ts
Adamj1232: Good call! Added null check for preview above this block although it should always be there and if an error occurs is caught in the fetch

PR #1542, opened 2023-02-21
Create new package for TallyHo name change to Taho

Adamj1232:
### Description
The rename request `TallyHo` -> `Taho` came from the Tally team on a rebrand that has occurred

Updated wallet name, docs and readme-s to make note of this (I don't think creating a new package is justified at this point and would be confusing and a pain for DApps who currently support Tally/Taho)

Screen Shot 2023-02-21 at 11 00 47 AM
Screen Shot 2023-02-21 at 11 00 38 AM

Request from Taho team:
Screen Shot 2023-02-21 at 11 20 11 AM

### Checklist


taylorjdawson:
@Adamj1232 I think they may want the package name to change? 🤔 Should we double check? It would be simple enough I suppose and then on the old tallyho package we could show it is deprecated? Or even use [this](https://stackoverflow.com/questions/28371669/renaming-a-published-npm-module)?

Adamj1232:
> @Adamj1232 I think they may want the package name to change? 🤔 Should we double check? It would be simple enough I suppose and then on the old tallyho package we could show it is deprecated? Or even use [this](https://stackoverflow.com/questions/28371669/renaming-a-published-npm-module)?

@taylorjdawson Yeah that is a good question and we discussed it yesterday in the planning meeting. Im waiting to hear back from the Tally/Taho team on this so we will see what their response is. Their initial request was to just update the displayed wallet name, but its prob best to create a new package and deprecate Tally

aaronbarnardsound (approved):

leightkt (approved):
I think creating a new package makes sense. Looks good!

PR #1549, opened 2023-02-24
feat: Ledger connector with support for WalletConnect v2

hlopes-ledger:
### Description

We (Ledger) have a working connector which supports the Ledger browser Extension and WalletConnect v1. This PR adds support for WalletConnect v2, based on code from the WalletConnect connector and an updated Ledger Connect Kit library.

We've added new parameters to the connector

- walletConnectVersion: 1 | 2,
- enableDebugLogs?: boolean,
- projectId: string
- requiredChains?: string[] | number[]
- requiredMethods?: string[]
- optionalMethods?: string[]
- requiredEvents?: string[]
- optionalEvents?: string[]

### Checklist
- [ ] The version field in `package.json` of the package you have made changes in is incremented following [semantic versioning](https://semver.org/) and using alpha release tagging

Wasn't able to run the internal demo, tested with this custom example app, https://64905dc39ea33271d1d1a2f4--ledger-w3o-vite-demo.netlify.app/

- [ ] send transaction
- [x] switch chains
- [x] sign message
- [x] sign typed message
- [x] disconnect


packages/ledger/src/v2.ts
```diff
@@ -0,0 +1,288 @@
...
+  WalletInit,
+  EIP1193Provider,
+  ProviderAccounts,
+  createEIP1193Provider
+} from '@web3-onboard/common'
+import type {
+  EthereumProvider as LedgerEthereumProvider
+} from '@ledgerhq/connect-kit-loader'
+import { isHexString, LedgerOptionsWCv2 } from './index.js'
+import { JQueryStyleEventEmitter } from 'rxjs/internal/observable/fromEvent'
```
aaronbarnardsound, line 12:
```suggestion
import type { JQueryStyleEventEmitter } from 'rxjs/internal/observable/fromEvent'
```
This appears to be used just as a type, so I think we can just import the type here.

packages/ledger/src/v2.ts
```diff
@@ -0,0 +1,288 @@
...
+  'eth_signTypedData_v4'
+]
+
+function ledger(options?: LedgerOptionsWCv2): WalletInit {
+  const projectId = options?.walletConnectVersion == 2 ? options.projectId : undefined
+  if (!projectId) {
+    throw new Error(
+      'WalletConnect requires a projectId. Please visit https://cloud.walletconnect.com to get one.'
+    )
+  }
```
aaronbarnardsound, line 30:
```suggestion
  if (!options?.projectId) {
    throw new Error(
      'WalletConnect requires a projectId. Please visit https://cloud.walletconnect.com to get one.'
    )
  }
```

Since we know that this will only be called for version 2, can we simplify to just check for the `projectId`?

packages/ledger/src/v2.ts
```diff
@@ -0,0 +1,288 @@
...
+        // or WalletConnect
+        const instance = (await connectKit.getProvider()) as LedgerEthereumProvider
+
+        // return the Ledger Extension provider
+        if (
+          checkSupportResult.providerImplementation ===
+            SupportedProviderImplementations.LedgerConnect
+        ) {
+          return {
+            provider: createEIP1193Provider(instance)
```
aaronbarnardsound, line 84: Just checking that this is intended to be wrapped with the `createEIP1193Provider` method, since the ledger extension provider is not wrapped in the [v1 file here](https://github.com/blocknative/web3-onboard/pull/1549/files#diff-e410806c59fd896721912f63f406b86661d71a193ed324b6f6f250e6556791e7R48)?

aaronbarnardsound (changes requested):
Looks good, just a couple of small changes requested.

I still need to test it out locally, but I wanted to ask which platforms is the Ledger Connect not available so that I can also try out the WalletConnect functionality?

hlopes-ledger:
> Looks good, just a couple of small changes requested.
> 
> I still need to test it out locally, but I wanted to ask which platforms is the Ledger Connect not available so that I can also try out the WalletConnect functionality?

Sorry for not including that on the PR description. You can test the WalletConnect support on Chrome; the Extension is currently only available on Safari iOS and macOS.

packages/ledger/src/v2.ts
hlopes-ledger: Yes, fixed. This was taken from the WalletConnect connector code, so I just wanted to maintain consistency with it.

packages/ledger/src/v2.ts
hlopes-ledger: Yes, fixed.

packages/ledger/src/v2.ts
hlopes-ledger: Well spotted, updated and tested it.

hlopes-ledger:
We've released the final version of Connect Kit and I've updated the package dependency.

leightkt (changes requested):
Please update the package.json for packages/ledger to: 
"version": "2.4.7-alpha.1",

packages/ledger/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/ledger",
-  "version": "2.4.6",
+  "version": "2.4.7-alpha.1",
```
Adamj1232, line 3:
```suggestion
  "version": "2.5.0-alpha.1",
```

packages/ledger/src/index.ts
```diff
@@ -1,279 +1,45 @@
...
+}
 
-        return {
-          provider: new EthProvider({ chains, connector })
-        }
-      }
-    }
-  }
+function ledger(options?: LedgerOptions): WalletInit {
+  const walletConnectVersion = options?.walletConnectVersion || 1
```
Adamj1232, line 39:
```suggestion
  const walletConnectVersion = options?.walletConnectVersion || 2
```
Defaults to version 2 of wallet connect

Adamj1232 (approved):
Nice work!

Issue #1520, opened 2023-02-09
[Feature]: allow instantiation of the web3-onboard with only the `ChainID` (no RPC URL, token name, etc) 


MOZGIII:
### Is your request related to a problem?

We are building the UI that doesn't know what chain it will operate on. We can't specify the RPC URL and/or chain name or token.

### Feature Description

Allow instantiating the `Onboard` without the chain parameters (besides `chainId`) known upfront. This is meant to support both public chains and private chains via the same deployment.

```ts
const chains = [
  { // no `token`, `label` or `rpcUrl`
    id: userSpecifiedChainId
  }
];

const onboard = Onboard({
  wallets,
  chains,
  appMetadata
})
```

The injected wallets are perfectly capable of carrying the sending of transactions without us specifying the RPC URL or other chain info. Further, wallet's `switchNetwork` should only require `chainId`, and the rest is only needed for adding a chain to the wallet. In our case this is a limiting factor.

### Alternative Solutions

We could display the form to the user to input the rest of the parameters, but that would mean that user has to enter the parameters required *only* for adding the chain to the wallet even if they already have the required chain in their wallet. In other words, it would just suck.

taylorjdawson:
Hey @MOZGIII! Good idea 💡  We will add this to our backlog and keep you posted here, thanks!

leightkt:
@MOZGIII Working on this currently. Finding that some wallets (metamask for example) require an rpcUrl to add a new chain to a wallet (switching between chains working fine with just an ID), and so a user MAY have to manually add a network that doesn't already exist in their wallet IF instantiated with only the chainId. Was this the intended behavior?

MOZGIII:
Yes, this is intended!

It would also be a nice to have for those wallets to allow passing their "extra" required parameters in the relevant place in their API, like when the wallet itself (not the whole `web3-onboard`) is configured. I assume the current design was to simplify the interaction, but it is actually limiting and confusing for the cases like this. It feels like doing things more explicitly would be so much better!

leightkt:
@MOZGIII PR up: https://github.com/blocknative/web3-onboard/pull/1589/ 
Does this meet your expectations or did you have other thoughts. Let me know! Thx!

PR #1589, opened 2023-03-14
[common, core] Update: allow instantiation with only chain id

leightkt:
### Description
Issue: https://github.com/blocknative/web3-onboard/issues/1520

Allows a new web-3-onboard instance to be instantiated with only an id required in the Chain object. 
Updates `setChain` function to accept optional arguments of rpcUrl, label, and token to add if needed to add a chain and chain was only instantiated with an id. 

Test Notes: 

In the Onboard object, set a chain or two (I used Goerli and Optimism) to only their Ids:
`{ id: '0x5' },`

`{ id: 10 },`
  
Test switching chains. 

after testing that initial behavior- add optional arguments to Optimism button and retest switch chains: 
`<button on:click={() => onboard.setChain({ chainId: 10, rpcUrl: 'https://mainnet.optimism.io', label: 'Optimism', token: 'OETH'})}
            >Set Chain to Optimism</button
          >`
          

### Checklist
- [X] Increment the version field in `package.json` of the package you have made changes in following [semantic versioning](https://semver.org/) and using alpha release tagging
- [X] Check the box that allows repo maintainers to update this PR
- [X] Test locally to make sure this feature/fix works
- [X] Run `yarn check-all` to confirm there are not any associated errors
- [X] Add or update relevant information in the documentation

### Docs Checklist
- [X] Include a screenshot of any changes ([see docs README on running locally](https://github.com/blocknative/web3-onboard/blob/develop/docs/README.md))
- [X] Add/update the appropriate package README (if applicable)
- [NA ] Add/update the related module in the [docs demo](https://github.com/blocknative/web3-onboard/blob/develop/docs/src/lib/services/onboard.js) (if applicable)
- [X] Add/update the related package in the `docs/package.json` file (if applicable)

- [X] send transaction
- [X] switch chains
- [X] sign message
- [X] sign typed message
- [X] disconnect

- [x] send transaction
- [x] switch chains
- [x] sign message
- [x] sign typed message
- [x] disconnect

Docs Changes:
Screenshot 2023-03-14 at 1 57 42 PM

To test changes: start the svelte project demo (yarn && yarn dev from root directory). Goerli and Optimism Chains have been instantiated with just a chainId. 

Video of Testing Changes:
[LOOM VIDEO](https://www.loom.com/share/42f12dee91c541de88cbc94cc4dd74d3)


packages/core/src/views/shared/NetworkSelector.svelte
```diff
@@ -6,6 +6,7 @@
   import { wallets$ } from '../../streams.js'
   import { distinctUntilChanged, debounceTime, skip } from 'rxjs/operators'
   import caretIcon from '../../icons/caret.js'
+  import { id } from 'ethers/lib/utils.js'
```
Adamj1232, line 9:
```suggestion
```

MOZGIII (changes requested):
Looks good to me! Although, not sure demo app changes are intended...

packages/core/src/views/shared/NetworkSelector.svelte
```diff
@@ -134,7 +135,7 @@
         >
       {/if}
       {#each chains as chain (chain.id)}
-        <option value={chain.id}>{chain.label}</option>
+        <option value={chain.id}>{chain.label || chainIdToLabel[chain.id]}</option>
```
Adamj1232, line 138: If the chain isn't in the natively supported chains do we want to display something? Maybe just chainId?

packages/demo/src/App.svelte
```diff
@@ -206,11 +206,14 @@
...
+      //   token: 'ETH',
+      //   label: 'Goerli',
+      //   rpcUrl: `https://goerli.infura.io/v3/${infura_key}`
+      // },
       {
-        id: '0x5',
-        token: 'ETH',
-        label: 'Goerli',
-        rpcUrl: `https://goerli.infura.io/v3/${infura_key}`
+        id: '0x5'
```
Adamj1232, line 216: Good for testing but prob want to revert after

docs/src/routes/docs/[...3]modules/core.md
```diff
@@ -113,10 +113,10 @@ An array of Chains that your app supports:
 ```ts
 type Chain = {
   id: ChainId // hex encoded string, eg '0x1' for Ethereum Mainnet
-  namespace?: 'evm' // string indicating chain namespace. Defaults to 'evm' but will allow other chain namespaces in the future
-  rpcUrl: string // used for network requests
-  label: string // used for display, eg Ethereum Mainnet
-  token: TokenSymbol // the native token symbol, eg ETH, BNB, MATIC
+  namespace?: 'evm' // string indicating chain namespace. Defaults to 'evm' but will allow other chain namespaces in the future 
+  rpcUrl?: string // Recommended to include. Used for network requests.
```
Comment (Adamj1232) [inline docs/src/routes/docs/[...3]modules/core.md:117]: Maybe include something around - some wallets require these for full functionality?

MOZGIII:
Just noting a good accompanying feature is to allow RPC URL to be passed explicitly in the appropriate places: i.e. when adding the chain to an injected wallet the RPC URL is required - but it's only *then*, and *if* it's needed.

---

I don't actually see the changes in this PR that would handle the chain addition without the URL. Will the wallet just return an error here?

---

@sesh92, could please provide a review here?

MOZGIII:
I'd expect some changes around here:

https://github.com/blocknative/web3-onboard/blob/5dcf0c937009b14e01f70d33893fa64d735e664b/packages/core/src/chain.ts#L76

To either allow passing the whole `chain` to add (when it's not in the init config), rather than `chainId`, or to handle the case where the RPC URL is not provided. Or is the latter not necessary because we'll be just getting an error, which is already handled?

leightkt:
@MOZGIII Sounds like you're looking for allowing for the passing of a rpcUrl (and token and label) as optional parameters in [setChain ](https://onboard.blocknative.com/docs/modules/core#setting-the-user-s-chain-network) i.e.:
`onboard.setChain({ chainId: 10, rpcUrl: 'https://mainnet.optimism.io', label: 'Optimism', token: 'OETH'})` to pass any needed parameters if they weren't set on instantiation. Is that correct? 
Can you provide some use case details? It will help me to better understand the end goal.

MOZGIII:
> end goal.

No goal in particular except for the better alignment of the internally available APIs and what they allow/offer with the potential use cases.

Currently, the `setChain` can't be used unless the chain was fully specified during the core initialization. It makes it impossible to add just the Chain ID to the chains list and init the core, and then, *if and when* the chain addition is required, dynamically obtain the parameters to add the chain (i.e. interactively from the user, or from a backend of some sort) - and then use them during `setChain`.

Ultimately, the way things would be set up after this PR, it would be possible to omit the parameters that are only required for `addChain` from the core initialization, which is good. It enables us to just use custom code for adding/switching the chain if needed, and doesn't force us to provide dummy values at core init.

I'd even put the newly optional values into a subobject to be more explicit about their nature and separate them from the rest (i.e. `chainId` is actually the only essential parameter, and the rest can be placed under `addChainParams?: AddChainParams`, where all `AddChainParams` fields are non-optional).

I.e. like:

```ts
type ChainId = number | biting | ...;
type Chain = {
  chainId: ChainId;
  addChainParams?: AddChainParams;
}
type AddChainParams = {
  rpcUrl: string;
  label: string;
  token: string;
  // ...
}
```

This way the API users know what the parameters are for, and the implementors know for what purpose the RPC URL is supposed to be used (i.e. don't accidentally start calling it directly instead of passing it to the wallet).

packages/core/src/provider.ts
```diff
@@ -445,20 +445,23 @@ export function switchChain(
...
+  label?: string,
+  token?: string,
 ): Promise<unknown> {
   return provider.request({
     method: 'wallet_addEthereumChain',
     params: [
       {
         chainId: chain.id,
-        chainName: chain.label,
+        chainName: chain.label || label,
```
Adamj1232, line 458: What happens if label, token or rpcUrl aren't defined on init and one or any subset are missing when this is called?

MOZGIII: In theory, the call should fail, and the provider should throw an implementation-dependent error. Not sure what will happen in practice, and how it's handled here...

leightkt: correct- the call will fail. We currently throw and error and display this message: 

Screenshot 2023-03-15 at 4 54 02 PM

My understanding is that it would be up to the dapp to handle?

MOZGIII: Yep. This modal is good enough for the case where the app doesn't care - if it does care, I think the whole `addChain` invocation logic can be implemented on the app side - to then follow with the onboard switch chain call. This is not the most straightforward way around the use of the available API for the users, but a good one for those who know what they're doing. We might want to note it in the docs though.

aaronbarnardsound (approved):
Just a small comment on a possible alternate solution for handling the updated chain values, but is really just an implementation detail that is not required for merge :)

packages/core/src/chain.ts
```diff
@@ -20,7 +23,7 @@ async function setChain(options: {
   }
 
   const { wallets, chains } = state.get()
-  const { chainId, chainNamespace = 'evm', wallet: walletToSet } = options
+  const { chainId, chainNamespace = 'evm', wallet: walletToSet, rpcUrl, label, token } = options
```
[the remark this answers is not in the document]
Adamj1232: Agree with keeping chain state aligned,  @leightkt's original approach 💰

packages/core/src/chain.ts
```diff
@@ -74,14 +75,19 @@ async function setChain(options: {
       code === ProviderRpcErrorCode.UNRECOGNIZED_CHAIN_ID
     ) {
       // chain has not been added to wallet
+      if (rpcUrl || label || token) {
+        // If values passed, update chain in state
+        chain.rpcUrl = rpcUrl
+        chain.label = label
+        chain.token = token
+        updateChain(chain)
+      }
```
aaronbarnardsound, line 84: This could be problematic if say the `rpcUrl` was set on init, but then not included in the call to`setChain`, it would get overwritten with `undefined`.
We might to just need to be a bit verbose and check if each value exists before assigning it.

```suggestion
      if (rpcUrl) {
        chain.rpcUrl = rpcUrl
      }

      if (label) {
        chain.label = label
      }
      
      if (token) {
        chain.token = token
      }
            
      updateChain(chain)
```

Another option would be to make `chain` a `let` var so we can reassign it and collect the additional chain update vars using the rest operator on line 27:

```javascript
const { chainId, chainNamespace = 'evm', wallet: walletToSet, ...chainUpdates } = options
```

and then merge them with `chain`

```javascript
if (chainUpdates) {
  chain = { ...chain, ...chainUpdates }
  updateChain(chain)
}
```

If we went this route, we'd just need to be careful if we add parameters to the `setChain` options as they would get merged in as well.

aaronbarnardsound (changes requested):
Just a small suggested change.

packages/core/src/chain.ts
leightkt: ah good catch! Thanks for the suggestion!

packages/core/src/chain.ts
```diff
@@ -71,11 +75,28 @@ async function setChain(options: {
...
+
+      // add chain to wallet
       return chainNotInWallet(
         wallet,
         chain,
         switchChainModalClosed$,
-        chainIdHex
+        chainIdHex,
```
Adamj1232, line 99: Has this file been prettier-ified? Trailing comma and line 27 looks a little long. No worries if it has just checking 😄

packages/core/src/provider.ts
```diff
@@ -445,7 +445,7 @@ export function switchChain(
 
 export function addNewChain(
   provider: EIP1193Provider,
-  chain: Chain
+  chain: Chain,
```
Adamj1232, line 448:
```suggestion
  chain: Chain
```

packages/core/src/store/actions.ts
```diff
@@ -71,13 +74,35 @@ export function addChains(chains: Chain[]): void {
...
       id: id.toLowerCase(),
-      rpcUrl: rpcUrl.trim()
+      rpcUrl: rpcUrl ? rpcUrl.trim() : null
     }))
   }
 
   dispatch(action as AddChainsAction)
 }
 
+export function updateChain(updatedChain: Chain): void {
```
Adamj1232, line 84: Is documentation for this function added to the core README and docs markdown?
There is a section around api functions

packages/core/src/views/shared/NetworkSelector.svelte
```diff
@@ -134,7 +134,7 @@
         >
       {/if}
       {#each chains as chain (chain.id)}
-        <option value={chain.id}>{chain.label}</option>
+        <option value={chain.id}>{chain.label || chainIdToLabel[chain.id] || chain.id}</option>
```
Adamj1232, line 137: Prettier?

packages/core/src/store/actions.ts
Adamj1232: We will also want to add a wrapper around this to the React package, maybe the Vue as well and then add documentation for it to eh React README along with the React docs md

packages/common/src/types.ts
```diff
@@ -386,9 +386,9 @@ export enum ProviderRpcErrorCode {
 export interface Chain {
   namespace?: 'evm'
   id: ChainId
-  rpcUrl: string
-  label: string
-  token: TokenSymbol // eg ETH, BNB, MATIC
+  rpcUrl?: string
```
Adamj1232, line 389: It would be helpful to add the notes for these props (and any missing) found in the README to the types as ts notes so they show up on hover in the IDE
EX:
```suggestion
  /**
   * Recommended to include. Used for network requests (eg Alchemy or Infura end point).
   * PLEASE NOTE: Some wallets require an rpcUrl, label, 
   * and token for actions such as adding a new chain. 
   * It is recommended to include rpcUrl, label, and token for full functionality. 
   */
  rpcUrl?: string
```
...or whatever fits within the prettier line-length max

packages/core/src/store/actions.ts
leightkt: @Adamj1232 does this function need to be separately documented in the core README? I've updated the documentation around the `setChain` function to include it will not accept an rpcUrl, token, and label. 
`updateChain` is currently used as part of the `setChain` function. 
`updateChain` isn't currently an exported function. Should I change this to make it available on it's own as part of the package, instead of just called inside the `setChain` function?

Adamj1232: Ohhh ahh you are correct then! All good!! No need to expose unless requested or needed. More coffee required on my end lol

packages/core/README.md
```diff
@@ -82,9 +82,11 @@ An array of Chains that your app supports:
 type Chain = {
   id: ChainId // hex encoded string, eg '0x1' for Ethereum Mainnet
   namespace?: 'evm' // string indicating chain namespace. Defaults to 'evm' but will allow other chain namespaces in the future
-  rpcUrl: string // used for network requests (eg Alchemy or Infura end point)
-  label: string // used for display, eg Ethereum Mainnet
-  token: TokenSymbol // the native token symbol, eg ETH, BNB, MATIC
+  // PLEASE NOTE: Some wallets require an rpcUrl, label, and token for actions such as adding a new chain. 
+  // It is recommended to include rpcUrl, label, and token for full functionality. 
+  rpcUrl?: string // Recommended to include. Used for network requests (eg Alchemy or Infura end point).
```
Adamj1232, line 87: It would be helpful to add all these prop notes as TS notes within `packages/common/src/types.ts` 
Def will enhance devEx while we are working in this area

docs/src/routes/docs/[...3]modules/react.md
```diff
@@ -196,7 +196,10 @@ type UseSetChain = (
 type SetChainOptions = {
   chainId: string
   chainNamespace?: string
-  wallet?: WalletState['label']
+  wallet?: WalletState['label'],
```
Comment (Adamj1232) [inline docs/src/routes/docs/[...3]modules/react.md:199]: Might be helpful to include the type notes to this for better understanding of prop usage here

aaronbarnardsound (approved):
Nice work, looks good!

PR #1597, opened 2023-03-17
Feature - New and Dynamic loading fonts

Adamj1232:
### Description
Add new font and dynamic import of them

### Checklist
- [ ] Increment the version field in `package.json` of the package you have made changes in following [semantic versioning](https://semver.org/) and using alpha release tagging
- [ ] Check the box that allows repo maintainers to update this PR
- [ ] Test locally to make sure this feature/fix works
- [ ] Run `yarn check-all` to confirm there are not any associated errors
- [ ] Confirm this PR passes Circle CI checks
- [ ] Add or update relevant information in the documentation


packages/core/src/index.ts
```diff
@@ -252,14 +265,18 @@ function mountApp() {
     customElements.define('onboard-v2', Onboard)
   }
 
-  // Add Fonts to main page
-  const styleEl = document.createElement('style')
+  if (!fontFamilyExternallyDefined(theme)) {
+    console.log('download font')
```
aaronbarnardsound, line 269:
```suggestion
```

packages/core/src/index.ts
```diff
@@ -241,7 +241,20 @@ function init(options: InitOptions): OnboardAPI {
   return API
 }
 
-function mountApp() {
+const fontFamilyExternallyDefined = (theme: Theme): boolean => {
+  if (!theme) return false
+  if (typeof theme === 'object' && theme['--w3o-font-family']) return true
+  if (
```
aaronbarnardsound, line 247: Based on this logic, if you did not pass in a `theme` object, but defined a CSS var for the font, it would still return `false`?

aaronbarnardsound (changes requested):
Just a console log to be removed and a question around the check if a font has been defined.

Also where is the dynamic loading occurring? It all looks like it is imported regardless of configuration from what I can tell?

Adamj1232:
> 

Refactored and moved to the CDN approach as recommended.
Dynamic importing can be seen in a few places, within Core it is now [here](https://github.com/blocknative/web3-onboard/pull/1597/files#diff-9a4ceebe7c6f86856371906c3f061d3b56b7457022b05179884a113e7ced67e8R263)

aaronbarnardsound (approved):
Looks good!

leightkt (approved):
Tested in the demo with useWebFont: true and passing custom font via --w3o-font-family

PR #1603, opened 2023-03-23
[core, react, vue]: Feature - Connect all previously connected wallets

Adamj1232:
### Description
Adds option to reconnect all previously connected wallets as an option.

Wallets will be reconnected in the order they were previously held within the array.

A new format is introduced within the PR for local storage saving of the wallet labels with an array of strings being used instead of a string. This will be the case regardless if the dev elects usage of `autoConnectAllPreviousWallet` or `autoConnectLastWallet`

```ts
type ConnectModalOptions = {
  ...
  /** 
   * If set to true, all previously connected wallets will store in 
   * local storage. Then on init, onboard will try to reconnect to 
   * each wallet with no modals displayed
   */
  autoConnectAllPreviousWallet?: boolean
}
```

### Checklist

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -231,22 +231,28 @@ function init(options: InitOptions): OnboardAPI {
     connect &&
     (connect.autoConnectLastWallet || connect.autoConnectAllPreviousWallet)
   ) {
-    const lastConnectedWallets = JSON.parse(
-      getLocalStore(STORAGE_KEYS.LAST_CONNECTED_WALLET)
+    const lastConnectedWallets = getLocalStore(
+      STORAGE_KEYS.LAST_CONNECTED_WALLET
     )
-
-    // Handle for legacy single wallet approach
-    if (lastConnectedWallets && typeof lastConnectedWallets === 'string') {
-      API.connectWallet({
-        autoSelect: { label: lastConnectedWallets, disableModals: true }
-      })
-    }
-    if (
-      lastConnectedWallets &&
-      Array.isArray(lastConnectedWallets) &&
-      lastConnectedWallets.length
-    ) {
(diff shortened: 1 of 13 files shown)


packages/core/src/index.ts
```diff
@@ -227,20 +227,59 @@ function init(options: InitOptions): OnboardAPI {
...
 
-    lastConnectedWallet &&
+    // Handle for legacy single wallet approach
+    if (lastConnectedWallets && typeof lastConnectedWallets === 'string') {
       API.connectWallet({
-        autoSelect: { label: lastConnectedWallet, disableModals: true }
+        autoSelect: { label: lastConnectedWallets, disableModals: true }
       })
+    }
+    if (lastConnectedWallets && Array.isArray(lastConnectedWallets)) {
```
aaronbarnardsound, line 244:
```suggestion
    if (lastConnectedWallets && Array.isArray(lastConnectedWallets && lastConnectedWallets.length)) {
```

Do we need to ensure there are wallets in the array here?

aaronbarnardsound (approved):
Just one question around checking the list lenght

packages/core/src/index.ts
Adamj1232: Ahh good call! Added!!

Adamj1232:
> When switching to this branch with a previously connected wallet in the demo project, I got this error: Screenshot 2023-03-23 at 5 00 01 PM
> 
> I can only load the demo app after I clear my local storage and reload.
> 
> Once I refreshed the page- I was able to test and worked as expected- reconnecting all wallets I had connected.

Ahh yep, forgot to add a try block to one area, should be fixed now for backwards compatibility throughout. Thanks for catching that!!

PR #1611, opened 2023-03-28
update favicon on docs site

leightkt:
### Description
Update favicon on docs site

Screenshot 2023-03-28 at 12 04 33 PM

### DIFF
--- a/docs/src/app.html
+++ b/docs/src/app.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


Adamj1232:
This does seem to fix the issue but I feel we are missing the sveltekit way for loading our favicon. The fact that we are getting a 404 on page load makes me think we are missing a config for static files. Maybe something [here](https://kit.svelte.dev/docs/configuration#files)?
If we choose this route should we remove the favicon within the project? cc @gesquinca

leightkt:
@Adamj1232 I was able to get it to load the existing favicon file by changing the path to `/favicon.png`

Adamj1232 (approved):
Nice!

PR #1690, opened 2023-05-09
Fix: Secondary Token Symbol fetch and docs

Adamj1232:
### Description
Updated api for `secondaryToken` by removing the token `name` attribute and instead fetch from the token contract for security measures and updated docs around this. Also, added a max length of 5 secondary tokens to validation along with amore error handling.

### Checklist


leightkt (approved):
Tested with USDC and DAI on the demo app

PR #1704, opened 2023-05-11
Feat: Private/custom rpc flow

Adamj1232:
### Description

This PR adds functionality to to Account Center expanded to  allow end users to update their RPC on mainnet.
If the user's primary wallet is on mainnet and `disableProtectedRpc` is not set to true within the w3o init the button will display to update their RPC. 
The RPC endpoint is configurable per chain obj (although this functionality is only enabled on mainnet) and will default to the BN protect RPC

There is also a localstorage set and get that adds a key after/if the user updates their RPC to the defined RPC so the btn will no longer be shown. We could update this to show a green shield in the future if protection is enabled.

```typescript
interface Chain {
  /** An optional protected RPC URL - Defaults to Blocknative's private and 
   * protected RPC to allow users to update the chain RPC within their wallet, 
   * specifically for private RPCs that protect user transactions
   */
  protectedRpcUrl?: string

  .....
}

type AccountCenter = {
  /**
   * false by default - This allows removal of the 
   * Enable Transaction Protection' button within the Account Center 
   * expanded when set to true
   * Can be set as a global for Account Center or per interface (desktop/mobile)
   */
  hideTransactionProtectionBtn?: boolean

  .....
}
```

Screenshot 2023-05-10 at 1 26 49 PM

### Checklist


packages/core/src/provider.ts
```diff
@@ -471,6 +471,31 @@ export function addNewChain(
   })
 }
 
+export function updateChainRPC(
+  provider: EIP1193Provider,
+  chain: Chain,
+  rpcUrl?: string
```
leightkt, line 477: should rpcUrl be required?

leightkt:
Button looks good and is working as expected!

packages/core/src/provider.ts
Adamj1232:
```suggestion
  rpcUrl: string
```

leightkt (approved):

packages/core/src/validation.ts
```diff
@@ -176,6 +176,7 @@ const accountCenter = Joi.object({
   position: commonPositions,
   expanded: Joi.boolean(),
   minimal: Joi.boolean(),
+  disableProtectedRpc: Joi.boolean(),
```
leightkt, line 179: I think you need to add disableProtectedRpc to accountCenterInitOptions too- I couldn't get it to initialize with disabledProtectedRpc: true otherwise

Adamj1232: Good catch! Forgot to commit those changes!

packages/core/src/index.ts
```diff
@@ -117,19 +122,22 @@ function init(options: InitOptions): OnboardAPI {
   if (typeof connect !== undefined) {
     updateConnectModal(connect)
   }
-
+  console.log(accountCenter)
```
leightkt, line 125:
```suggestion
```

leightkt (approved):
one console.log to remove- otherwise looking good and working as expected

aaronbarnardsound (approved):
Looks good! Just a small typo in there I think.

packages/core/src/views/connect/Index.svelte
```diff
@@ -343,6 +343,7 @@
     if (
       appChain &&
       !secondaryTokens &&
+      appChain &&
```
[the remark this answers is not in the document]
Adamj1232: Ahh good catch!

PR #1687, opened 2023-05-08
Update React Module and Quickstart Steps

leightkt:
### Description
Update react module and quickstart steps to be more accurate in documenation

### Docs Checklist
- [ ] Include a screenshot of any changes ([see docs README on running locally](https://github.com/blocknative/web3-onboard/blob/develop/docs/README.md))
- [ ] Add/update the appropriate package README (if applicable)
- [ ] Add/update the related module in the [docs demo](https://github.com/blocknative/web3-onboard/blob/develop/docs/src/lib/services/onboard.js) (if applicable)
- [ ] Add/update the related package in the `docs/package.json` file (if applicable)

Screenshot 2023-05-08 at 12 36 10 PM
Screenshot 2023-05-08 at 12 36 36 PM
Screenshot 2023-05-08 at 12 36 51 PM
Screenshot 2023-05-08 at 12 37 04 PM

Screenshot 2023-05-08 at 12 37 20 PM

PR #1665, opened 2023-04-28
Enhancement: Docs upgrade- kit-docs and sveltejs/kit

leightkt:
### Description
Upgrades: 
@svelteness/kit-docs: 1.1.3
@sveltejs/kit: 1.5.0
Fixes font loading
Routes restructured to work with new kit-docs structure: [https://kitdocs.vercel.app/docs/getting-started/quickstart](https://kitdocs.vercel.app/docs/getting-started/quickstart)

### Checklist
- [NA] Increment the version field in `package.json` of the package you have made changes in following [semantic versioning](https://semver.org/) and using alpha release tagging
- [x] Check the box that allows repo maintainers to update this PR
- [x] Test locally to make sure this feature/fix works
- [x] Run `yarn check-all` to confirm there are not any associated errors
- [x] Confirm this PR passes Circle CI checks
- [x] Add or update relevant information in the documentation

PR #1705, opened 2023-05-11
Release 2.22.1 (docs)

leightkt:
#### Docs:

- Update: Upgrade Docs to @svelteness/kit-docs 1.1.3 #1665 
- Update: React Quickstart Examples #1687

### **📦 Changes per package**


docs/src/lib/services/onboard.js
```diff
@@ -106,6 +105,8 @@ const intiOnboard = async (theme) => {
     apiKey: 'pk_live_02207D744E81C2BA'
   })
 
+  const frame = frameModule()
+
   return Onboard({
     wallets: [
```
Adamj1232, line 111: Do we want to add Frame to the wallets list? If so prob a little lower in the lineup

Adamj1232: nm just saw it down there!

PR #1712, opened 2023-05-12
Patch: Fix docs build

leightkt:


docs/package.json
```diff
@@ -47,6 +47,7 @@
   },
   "type": "module",
   "dependencies": {
+    "@sveltejs/adapter-static": "^2.0.2",
```
Adamj1232, line 50: I think this should be a devDep and we can remove the adapter-auto

docs/svelte.config.js
```diff
@@ -1,4 +1,4 @@
-import adapter from '@sveltejs/adapter-auto'
+import adapter from '@sveltejs/adapter-static';
```
Adamj1232, line 1:
```suggestion
import adapter from '@sveltejs/adapter-static'
```

Adamj1232 (approved):

PR #1723, opened 2023-05-16
Enhancement: Account Center Updates

leightkt:
### Description
Update Account Center default location, animation, and position of components. 

[Loom video of testing](https://www.loom.com/share/daa381ef4aee4bdfa1b329dc5c1ea938)

Screenshot 2023-05-16 at 10 33 09 AM

Screenshot 2023-05-16 at 10 33 32 AM

### Checklist

(docs update for default position included in kit-docs upgrade)


Adamj1232 (commented):
Should AC be displaying as micro by default?
By adding a bit of z-index to the minimized AC the animation of the Maximized leaving looks a bit cleaner. Doesnt have to be z-index but I think we can clean up that animation a tidbit.

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/core",
-  "version": "2.18.1-alpha.2",
+  "version": "2.18.1-alpha.3",
```
Adamj1232, line 3: Since this is changing the default way core displays I am thinking we should make this a minor version bump

Adamj1232:
> @Adamj1232 I wouldn't classify this as a breaking change- things will still load, the default location has just changed. But yes we can add instructions for how to reset back to the default location on the release- probably a good idea. 

Sweet, I think it will be helpful! Breaking changes in ui lib development would also include changes to design systems. The default position to a component would fall under this. As most uses of AC are in the default position this change would break the design of those apps.

leightkt:
> > @Adamj1232 I wouldn't classify this as a breaking change- things will still load, the default location has just changed. But yes we can add instructions for how to reset back to the default location on the release- probably a good idea.
> 
> Sweet, I think it will be helpful! Agreed it won't break the build - unless the have you test for position or layout but breaking changes in ui lib development would also include changes to design systems. The default position to a component would fall under this. As most uses of AC are in the default position this change would break the design of those apps.

Ahhhhh Super helpful to know! I was thinking of a breaking change in terms of a backend library... but that makes total sense!

gesquinca:
> Should AC be displaying as micro by default?

@Adamj1232 yes, this is the plan going forward. 

> As this is going to change the default behavior for dapps who use account center currently - which could be frustrating - how can we note this change is breaking...Maybe something big on the top of the readme for this specific release(or maybe the next few) with steps to get back to the previous default behavior for AccountCenter (just a code snip of init options for relocating). Any ideas @leightkt @aaronbarnardsound @gesquinca ?

@Adamj1232 100% on this. These changes should be accompanied by some documentation so users can display account center in the old way if they wish to do so.

gesquinca:
@leightkt Seems like we lost the ability to close Account Center panel when clicking outside. We definitely want to keep that.  Could you double-check this is the case?

leightkt:
@Adamj1232 Included a warning at the top of the core readme and the core module page in docs. I figure after a month or so we can move further down the page. 

Screenshot 2023-05-18 at 11 09 53 AM

Adamj1232 (approved):
Looks good! Nice work!!

PR #1779, opened 2023-06-15
Feature- update appMetadata

leightkt:
### Description
Allow for the updating of appMetadata after initialization

### Checklist

### Docs Checklist
- [x] Include a screenshot of any changes ([see docs README on running locally](https://github.com/blocknative/web3-onboard/blob/develop/docs/README.md))
- [x] Add/update the appropriate package README (if applicable)
- [NA] Add/update the related module in the [docs demo](https://github.com/blocknative/web3-onboard/blob/develop/docs/src/lib/services/onboard.js) (if applicable)
- [x] Add/update the related package in the `docs/package.json` file (if applicable)

Screenshot 2023-06-14 at 8 30 45 PM


packages/core/src/index.ts
```diff
@@ -217,6 +218,8 @@ function init(options: InitOptions): OnboardAPI {
     unstoppableResolution
   })
 
+  updateAppMetadata(appMetadata)
```
Adamj1232, line 221: I think we need an undefined check here and remove the default null initializer on ln 95

leightkt: updated!

packages/core/src/index.ts
```diff
@@ -217,6 +218,8 @@ function init(options: InitOptions): OnboardAPI {
     unstoppableResolution
   })
 
+  appMetadata ? updateAppMetadata(appMetadata) : null
```
Adamj1232, line 221:
```suggestion
  appMetadata && updateAppMetadata(appMetadata)
```

Adamj1232: This is a little more clean than a ternary that resolves to null. Resolving to null would make more sense if the ternary was used to set a variable.
A basic if block would be fine also

packages/demo/src/App.svelte
```diff
@@ -307,23 +307,23 @@
...
-      ],
-      agreement: {
-        version: '1.0.0',
-        termsUrl: 'https://www.blocknative.com/terms-conditions',
-        privacyUrl: 'https://www.blocknative.com/privacy-policy'
-      },
-      gettingStartedGuide: 'https://blocknative.com',
-      explore: 'https://blocknative.com'
-    },
+    // appMetadata: {
```
Adamj1232, line 310: This can be commented back in

Adamj1232 (approved):
Nice!!

PR #1797, opened 2023-06-28
Fix: Ledger Walletconnect v2 handling and build

Adamj1232:
### Description
Fix import breaking build
Remove log from WalletConnect package


packages/demo/src/App.svelte
```diff
@@ -472,15 +472,21 @@
       )
       verifySign = recoveredAddress === addr
     } catch (error) {
-      console.error('Error recovering addressL', error);
+      console.error('Error recovering addressL', error)
```
leightkt, line 475:
```suggestion
      console.error('Error recovering address', error)
```

leightkt (approved):

leightkt:
@Adamj1232 do we need to update the ledger module in the docs with our project id as well?

taylorjdawson (approved):

Adamj1232:
> @Adamj1232 do we need to update the ledger module in the docs with our project id as well?

Ah good catch!1!

PR #2075, opened 2024-03-04
remove deprecated module

BitgetWalletTeam:
https://onboard.blocknative.com/docs/wallets/bitkeep
hi, our new brand name is Bitget Wallet, and 2 modules in this repo sametime, for user experience we need to remove the old one, please confirm and accept this request, thanks.

from: Bitget Wallet Team


packages/bitkeep/src/index.ts
```diff
@@ -1,54 +0,0 @@
...
-
-/**
- * @deprecated Please use `@web3-onboard/bitget` instead.
- */
-function bitKeep(): WalletInit {
-  if (typeof window === 'undefined') return () => null
-
-  return () => {
-    return {
-      label: 'BitKeep',
```
Adamj1232, line 28: Rather than removing this module, updating the displayed name is more simple for dapp devs to handle maintenance and follows our process. We can add a deprecation flag but we wont want to remove this from the repo

Adamj1232:
@BitgetWalletTeam friendly ping on how to proceed? Marking the module as deprecated is the simplest along with a name update to the module for the wallet displayed name returned would get the desired results with backwards compatibility

Adamj1232:
Closing PR as it has become stable and utilizes an approach that doesn't align with our deprecation processes

PR #2188, opened 2024-05-06
Integration of FinoaConnect wallet connector service into Web3Onboard

finoaamukherjee:
### Description
Integration of FinoaConnect (Finoa's Wallet Connector Service) into Web3Onboard for distribution through BlockNative. For any questions or queries, please reach out to:
* Anirban Mukherjee (anirban.mukherjee@finoa.io)
* Laura Pescatori (laura.pescatori@finoa.io)

### Checklist
- [X] Increment the version field in `package.json` of the package you have made changes in following [semantic versioning](https://semver.org/) and using alpha release tagging
- [X] Check the box that allows repo maintainers to update this PR
- [X] Test locally to make sure this feature/fix works
- [X] Run `yarn check-all` to confirm there are not any associated errors
- [ ] Confirm this PR passes Circle CI checks
- [X] Add or update relevant information in the documentation

### Docs Checklist
- [X] Include a screenshot of any changes ([see docs README on running locally](https://github.com/blocknative/web3-onboard/blob/develop/docs/README.md))
- [X] Add/update the appropriate package README (if applicable)
- [X] Add/update the related module in the [docs demo](https://github.com/blocknative/web3-onboard/blob/develop/docs/src/lib/services/onboard.js) (if applicable)
- [X] Add/update the related package in the `docs/package.json` file (if applicable)

### If this PR includes changes to add an injected wallet or SDK wallet module: 
Please complete the following using the internal demo package.
To run this demo use the command `yarn && yarn dev` to get the project running at `http://localhost:8080/`
[Finoa] It was verified that Finoa's documentation appears under `docs`. The `FinoaConnect` option does show up on the list of injected wallets, and we have verified that connecting and disconnecting the wallet to the dApp works fine.

#### Tests with demo app (injected)
- [ ] send transaction
- [ ] switch chains
- [ ] sign message
- [ ] sign typed message
- [X] disconnect

#### Tests with demo app (SDK)
- [x] send transaction
- [x] switch chains
- [x] sign message
- [x] sign typed message
- [X] disconnect

### Screenshots

Screenshot of FinoaConnect under `docs`
Screenshot 2024-05-06 at 14 48 06

PR #2193, opened 2024-05-14
feat: add passport wallet

k-xo:
### Description

[Passport](https://0xpass.io/) is an MPC-based programmable, distributed, and non-custodial key management system, that allows users to generate wallets, scoped to their application, either via user Passkeys, our signer allows you to sign messages and transactions with a Passport Network account.

### Checklist
- [ ] Confirm this PR passes Circle CI checks

### Docs Checklist
- [x] Include a screenshot of any changes ([see docs README on running locally](https://github.com/blocknative/web3-onboard/blob/develop/docs/README.md))
- [x] Add/update the appropriate package README (if applicable)
- [x] Add/update the related module in the [docs demo](https://github.com/blocknative/web3-onboard/blob/develop/docs/src/lib/services/onboard.js) (if applicable)
- [x] Add/update the related package in the `docs/package.json` file (if applicable)

- [ ] send transaction
- [ ] switch chains
- [ ] sign message
- [ ] sign typed message
- [ ] disconnect

- [x] send transaction
- [x] switch chains
- [x] sign message
- [x] sign typed message
- [x] disconnect

### DIFF
--- a/packages/demo/package.json
+++ b/packages/demo/package.json
@@ -47,7 +47,7 @@
     "@web3-onboard/keystone": "^2.3.8",
     "@web3-onboard/ledger": "^2.6.0",
     "@web3-onboard/magic": "^2.1.6",
-    "@web3-onboard/metamask": "^2.0.6",
+    "@web3-onboard/metamask": "^2.0.7-alpha.1",
     "@web3-onboard/mew-wallet": "^2.0.3",
     "@web3-onboard/phantom": "^2.0.3",
     "@web3-onboard/portis": "^2.1.6",
--- a/docs/package.json
+++ b/docs/package.json
@@ -72,7 +72,7 @@
     "@web3-onboard/keystone": "^2.3.7",
     "@web3-onboard/ledger": "^2.6.0",
     "@web3-onboard/magic": "^2.1.7",
-    "@web3-onboard/metamask": "^2.0.6",
+    "@web3-onboard/metamask": "^2.0.7-alpha.1",
     "@web3-onboard/mew-wallet": "^2.0.4",
     "@web3-onboard/phantom": "^2.0.3",
     "@web3-onboard/portis": "^2.1.7",
--- a/packages/metamask/package.json
+++ b/packages/metamask/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@web3-onboard/metamask",
-  "version": "2.0.6",
+  "version": "2.0.7-alpha-1",
   "description": "MetaMask SDK wallet module for connecting to Web3-Onboard. Web3-Onboard makes it simple to connect Ethereum hardware and software wallets to your dapp. Features standardised spec compliant web3 providers for all supported wallets, framework agnostic modern javascript UI with code splitting, CSS customization, multi-chain and multi-account support, reactive wallet state subscriptions and real-time transaction state change notifications.",
   "keywords": [
     "Ethereum",
@@ -60,7 +60,7 @@
     "typescript": "^5.2.2"
   },
   "dependencies": {
-    "@metamask/sdk": "^0.18.6",
+    "@metamask/sdk": "^0.20.3",
     "@web3-onboard/common": "^2.3.3"
   },
   "engines": {
--- a/packages/metamask/src/index.ts
+++ b/packages/metamask/src/index.ts
@@ -42,7 +42,7 @@ function metamask({
       label: 'MetaMask',
       getIcon: async () => (await import('./icon.js')).default,
       getInterface: async ({ appMetadata }) => {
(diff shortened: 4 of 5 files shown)


packages/demo/src/App.svelte
```diff
@@ -450,6 +467,7 @@
   const signTransactionMessage = async provider => {
     // if using ethers v6 this is:
     // ethersProvider = new ethers.BrowserProvider(wallet.provider, 'any')
+    console.log('here here')
```
Adamj1232, line 470:
```suggestion
```

packages/passport/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@web3-onboard/passport",
+  "version": "2.0.0",
```
Adamj1232, line 3:
```suggestion
  "version": "2.0.0-alpha.1",
```

packages/demo/package.json
```diff
@@ -64,6 +64,9 @@
     "@web3-onboard/xdefi": "^2.0.5",
     "@web3-onboard/zeal": "^2.0.3",
     "@web3-onboard/particle-network": "^2.0.1",
+    "@web3-onboard/passport": "2.0.0",
+    "@0xpass/webauthn-signer": "2.0.0",
+    "@0xpass/passport": "2.1.1",
```
Adamj1232, line 69: Are these deps required?

packages/passport/yarn.lock
```diff
@@ -0,0 +1,1040 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Adamj1232, line 1: This file can be removed from the package level

packages/passport/src/index.ts
```diff
@@ -0,0 +1,296 @@
+import { createEIP1193Provider, type WalletInit } from '@web3-onboard/common'
+import { Network, Passport } from '@0xpass/passport'
+import { createPassportClient } from '@0xpass/passport-viem'
+import { http } from 'viem'
+import { mainnet } from 'viem/chains'
+import { AuthenticatedHeaders } from '@0xpass/passport/dist/types'
+import CryptoJS from 'crypto-js'
+import passportLogo from './passportlogo.svg'
+import { PassportOptions } from 'types'
```
Adamj1232, line 9: We will want any non devDependency to be imported dynamically so that imports are light on dapps on start and deps are only pulled in once passport is selected from the web3-onboard options.

Any types should also be imported as `import type {....} from '....'`

packages/demo/package.json
k-xo: Yes these deps are required

Adamj1232: Why are they required if they are part of the `@web3-onboard/passport` package?

packages/passport/src/index.ts
```diff
@@ -0,0 +1,298 @@
+import { createEIP1193Provider, type WalletInit } from '@web3-onboard/common'
```
Adamj1232, line 1:
```suggestion
import type { WalletInit } from '@web3-onboard/common'
```
Can you please split this out as well and dynamically import `createEIP1193Provider` within the `getInterface` method?

packages/passport/README.md
```diff
@@ -0,0 +1,84 @@
+# @web3-onboard/passport
```
Adamj1232, line 1: We will also need to add this info around the package to the internal docs site similar to as being done here - https://github.com/blocknative/web3-onboard/pull/2188/files#diff-880dac278a88e1be5c5d7e94dedea953ce24d29259488ff2280d1c754b5f4f1d

k-xo: added

packages/passport/src/index.ts
k-xo: done

packages/demo/package.json
k-xo: actually for this you're right passport can be removed but webauthn-signer is a required dependency.

Adamj1232: @k-xo if it is required for the package to work should it be added to the package dependencies?
If a dependency is required for the package to work and isnt included it should be documented

k-xo: We have multiple dependency types, and support a specific signer interface, so this setup is setting up the flow where users are using passkeys / webauthn signer, so it is required in the application context of setting up the signer, and is the most convenient way of setting it up, however it is not required in the sense it should be included in the onboard package. 

so in the demo we have the passport onboard package and the webauthn signer, but in the actual passport package we have the core passport sdk, as well as our viem client.

k-xo:
@Adamj1232 any pointer to fixing the build issue? haven't made any changes to any other packages, was working fine at a point, and now... isn't unsure why

PR #2204, opened 2024-05-28
Enh - Coinbase dependency upgrade to 4.0

Adamj1232:
### Description
Coinbase dependency upgrade to 4.0
This support the new Coinbase smartwallet usage

### Checklist

- [x] send transaction
- [x] switch chains
- [x] sign message
- [x] sign typed message
- [x] disconnect


packages/coinbase/src/index.ts
```diff
@@ -31,29 +45,42 @@ function coinbaseWallet({
...
         ) as typeof CoinbaseWalletSDK
+        const { isHex, toHex } = await import('@web3-onboard/common')
 
         const base64 = window.btoa(icon || '')
         const appLogoUrl = `data:image/svg+xml;base64,${base64}`
 
+        const appChainIds = chains.map(
+          // @ts-ignore - treating hex strings as numbers as they are expected to be hex numbers
+          ({ id }) => id as number
+        )
```
nateReiners, line 56: CoinbaseWalletSDK is expecting `appChainIds: number[]`, so this should be something like 

```
const appChainIds = chains.map(({ id }) => convertChainIdToNumber(id))
```

Adamj1232: @nateReiners the number type accepts hex numbers as does the coinbase SDK from my testing. Am I incorrect in this finding?
The SDK does work as expected when using hex IDs and W3O expects chainId in hex.

Adamj1232: Using https://coinbase.github.io/coinbase-wallet-sdk/ on requestAccounts connect returns a hex so its unclear why I would pass in an int number instead of hex number

Adamj1232: chainChanged also returns a hex number.
`({ id }) => id as number` the `id` here is a hex number not a string number

nateReiners: > number type accepts hex numbers

If this were true we wouldn't need any typecasting here right? 

> Am I incorrect in this finding

I think so. The CoinbaseWalletSDK uses the first chainId in `appChainIds` as the default chain for the next request you make after connecting, and a hex string won't be recognized by the Smart Wallet popup as a supported chain. Here's how you can test it: 

1. Run the demo project locally by running `yarn dev` from root and visiting localhost:8080
2. Connect with Coinbase Wallet
3. Click the 'Sign Message' button - this request will be made on whatever chainID is the first chain ID in the `appChainIds` array, in the case of the demo app it is `"0x1"`. 

What you'll see is an unsupported network error in the Smart Wallet Popup.

Made the change here to save you some time :)  https://github.com/blocknative/web3-onboard/pull/2212

Adamj1232: Thanks for that @nateReiners although W3O expects chain Ids returned from the sdk to be in hex format so passing in number values causes issues down the line in core and common. I am not able to catch a `connect` event to convert the value to for w3o. which event should fire upon connect?

Adamj1232: When using the eoa wallet the chainId comes through as hex

packages/coinbase/src/index.ts
```diff
@@ -31,29 +45,42 @@ function coinbaseWallet({
...
-              listener(`0x${(val as number).toString(16)}`)
+              let hexVal: string
+              if (isHex(val)) {
+                hexVal = val
+              } else {
+                hexVal = toHex(val)
+              }
+
+              // @ts-ignore
+              listener(hexVal)
```
nateReiners, line 83: Thank you for handling this. We will update the `chainChanged` event in the ext and mobile providers to make them EIP-1193 compliant. Appreciate the workaround in the meantime 🙏

packages/coinbase/src/index.ts
Adamj1232: @nateReiners can the SDK be updated to handle hex chainIds along with number values? Use of hex chain ID is pretty consistent throughout

nateReiners: > W3O expects chain Ids returned from the sdk to be in hex format so passing in number values causes issues down the line in core and common

Understood. To clarify, you mean the chainId returned from an `eth_chainId` request should be a hex but is a number instead? Thank you for raising this -- this looks like a bug in our provider which we will address ASAP. In the meantime you'll have to convert the eth_chainId response to hex on your end, but you can expect 4.0.3 with that fix very soon. Are there any other spots you are expecting a hex but we're returning a number, or is it just `eth_chainId` requests?

Other libs that use @coinbase/wallet-sdk like Viem and Wagmi have chain models where the id is represented as a number, so I get what you're saying about converting to number only to get hex back again, but either way someone will have to do a hex/number conversion so we're going to leave it as `appChainIds: number[]` for now.

> I am not able to catch a connect event to convert the value to for w3o

The provider does emit a 'connect' event with the chainID as a hex upon successful connection [LOC](https://github.com/coinbase/coinbase-wallet-sdk/blob/master/packages/wallet-sdk/src/CoinbaseWalletProvider.ts#L77). 

You should be able to listen for it like this 

```
const provider = sdk.makeWeb3Provider({options: 'smartWalletOnly'});

provider.on('connect', (info) => {
    setConnect(info);
});
```

If you can point me to the LOC I can help debug if that listener isn't working.

packages/coinbase/src/index.ts
```diff
@@ -62,9 +89,20 @@ function coinbaseWallet({
 
           return coinbaseWalletProvider
         }
+        const provider = createEIP1193Provider(coinbaseWalletProvider, {
```
Adamj1232, line 92: @nateReiners patched provider to handle number value returned from `eth_chainId` - just flagging for visibility 😄

packages/coinbase/src/index.ts
Adamj1232: Thanks for that @nateReiners ! I have a patched the provider for that request in the meantime and tagged you in it in the PR. Apologies for the lack of clarity above but I think we have this at a good spot for an alpha release. All other testing has yielded good results.
Thank you again for your help!

nateReiners: PR to fix `eth_chainId` result: https://github.com/coinbase/coinbase-wallet-sdk/pull/1310

nateReiners: @Adamj1232 Yep agree the branch looks good to me!

packages/coinbase/src/index.ts
nateReiners: @Adamj1232 If you want to bump the sdk to 4.0.3 and get rid of this workaround, the version with the `eth_chainId` fix was just published https://www.npmjs.com/package/@coinbase/wallet-sdk/v/4.0.3

nateReiners (approved):

packages/coinbase/src/index.ts
Adamj1232: @nateReiners when I remove the patch and `createEIP1193Provider` function and just use the `coinbaseWalletProvider` I get the following error after bumping to 4.0.3 - I dont get this error on 4.0.2 without the method. 
`eth_chainId` fix is working well!
Screenshot 2024-05-31 at 11 11 24

PR #2211, opened 2024-05-30
Update hw-wallet packages for type alignment

Adamj1232:
### Description
Update hw-wallet packages for type alignment


.github/workflows/docs.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Setup Nodejs and yarn
         uses: actions/setup-node@v2
         with:
-          node-version: "18.18"
+          node-version: "20"
```
Adamj1232, line 43:
```suggestion
          node-version: "18.18"
```

PR #2253, opened 2024-07-09
feat: add gate wallet

0xssw:


0xssw:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

Adamj1232:
As an injected wallet, with no SDK required they dont need to create their own module but rather only add their wallet to the injected module. Creating a separate module creates extra overhead for the team and all functionality is available through the injected module without dapps needing to add an additional package.

PR #2390, opened 2026-05-04
Migrate to @metamask/connect-evm

0xFirekeeper:
Swap out the legacy @metamask/sdk for the new @metamask/connect-evm client and update package metadata (version bump to 2.3.0). Code changes load connect-evm (handling ESM/CJS shapes), reuse a single client instance, attach a disconnect shim to the provider, and map legacy MetaMask SDK options to the new connect-evm option shape (including Infura/custom RPC merging via getInfuraRpcUrls). README updated to reflect the Connect EVM integration and backwards compatibility notes; package.json dependency updated accordingly.

## Summary by CodeRabbit

* **New Features**
  * Rebranded to MetaMask Connect EVM and prioritized when a MetaMask browser wallet is present.
  * Added optional config: headless, showInstallModal, infuraAPIKey, readonlyRPCMap, openDeeplink, useDeeplink; dappMetadata is now optional and supports iconUrl.
  * Legacy MetaMask options remain accepted for backwards compatibility.

* **Documentation**
  * Updated README and usage example to show Connect EVM integration and option mappings.

* **Chores**
  * Package version bumped to 2.3.0.

[Review Change Stack](https://app.coderabbit.ai/change-stack/thirdweb-dev/web3-onboard/pull/2390?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/metamask/README.md
+++ b/packages/metamask/README.md
@@ -1,8 +1,9 @@
 # @web3-onboard/metamask
 
-## Wallet module for connecting MetaMask Wallet SDK to web3-onboard
+## Wallet module for connecting MetaMask Connect EVM to web3-onboard
 The MetaMask Web3-Onboard module provides a reliable, secure, and seamless connection from your dapp to the MetaMask browser extension and MetaMask Mobile.
-See [MetaMask SDK Developer Docs](https://docs.metamask.io/wallet/how-to/connect/set-up-sdk/)
+
+This module uses [MetaMask Connect EVM](https://docs.metamask.io/metamask-connect/evm/) (`@metamask/connect-evm`) under the hood — the successor to the legacy `@metamask/sdk`. The integration surface for `@web3-onboard/metamask` is unchanged: the legacy options below are mapped to their MetaMask Connect EVM equivalents internally so existing dapps keep working without code changes.
 
 ![MetaMask SDK ConnectionFlow](https://github.com/blocknative/web3-onboard/blob/develop/assets/metaMaskSDK-connect.gif?raw=true 'MetaMask SDK ConnectionFlow')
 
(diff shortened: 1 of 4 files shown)


wenfix:
Hey, great work on the migration to our new SDK. I've pulled, tested it and found a couple issues:

1. Can you bump `@metamask/connect-evm` to `1.3.1` and regenerate the relevant `yarn.lock` files?

   `1.3.1` includes a [provider fix](https://github.com/MetaMask/connect-monorepo/pull/297) that Web3-Onboard needs for account requests. Web3-Onboard calls:

   ```ts
   provider.request({ method: 'eth_requestAccounts' })
   ```

   and expects the standard EIP-1193 response shape, which is an array of connected accounts. With `1.3.1`, intercepted account requests return the expected method-specific values: `eth_requestAccounts` returns the accounts array, and `eth_coinbase` returns the selected account.

2. I think the QR rendering issue is coming from the icon metadata being passed into Connect EVM.

   The current implementation appears to take Web3-Onboard's `appMetadata.icon`, base64 encode it, and pass it as `dapp.base64Icon`. In Web3-Onboard, `appMetadata.icon` can be a raw SVG string, and the demo app uses a fairly large inline SVG there.

   Connect EVM includes dapp metadata in the mobile connection request, which then gets compressed into the QR/deeplink payload. When the inline SVG is included as `base64Icon`, the payload gets too large and QR generation can fail with `Capacity overflow`.

   I would avoid automatically mapping `appMetadata.icon` to `base64Icon`. Safer options would be:

   - use an explicit `options.dappMetadata.iconUrl` when provided
   - only pass `options.dappMetadata.base64Icon` if the integrator explicitly provides it
   - if no explicit Connect EVM icon metadata is provided, omit the icon and let Connect fall back to favicon/default behavior
   - if `appMetadata.icon` is already an HTTP(S) URL, forwarding it as `iconUrl` seems reasonable, but raw SVG strings should not be inlined into the QR payload


Let me know if you have questions or need further help! 🙏

